PR #3, opened 2017-06-20
Set webpack targetLibrary to umd

lilfaf:
Allows any module definitions (CommonJS, AMD or global var). 
Congrats for the great job you've done on this lib ! Regards


lilfaf:
I'm not quite sure about the necessity to use yarn for the build script. Most projects I know remained with npm.

arnog:
But couldn't the yarn.lock file get out of date if, let's say, a new package is added via `npm install` by a non-yarn using user? Or will yarn detect the added package and update yarn.lock the next time a yarn users uses yarn?

PR #4, opened 2017-06-22
Updated node version

tom-s:
Prevent error "The engine "node" is incompatible with this module. Expected version "~8.0"." when using node v8.1 and above


arnog:
Good catch! I think the "engines" property could be removed altogether since the library is not, in fact, dependent on node.js.

PR #24, opened 2017-10-19
Correct logic for inlineShortcuts override. Fixes #23.

### LINKED ISSUE #23: inlineShortcuts to makeMathField doesn't override old values
## Description

If `inlineShortcuts` is given as an option to makeMathField, it should override any old values. It doesn't.

For example, if the value is
```
inlineShortcuts: { '>-': '>-',	// override builtin shortcut (\succ)
			   '<-': '<-'},	// override builtin shortcut (\leftarrow)
```
Then when `>-` is typed, the `\succ` symbol appears instead of the override `>-` (i.e, don't do the substitution).

NSoiffer:
If overriding a default, the string lengths will be equal, so the test should be `>=`, not `>`. Also, it seems one of the strings is created with 'new', so `==` works but  `===` doesn't. I didn't look into the details, but that tweak caused the comparison to be true.

### DIFF
--- a/src/addons/auto-render.js
+++ b/src/addons/auto-render.js
@@ -1,5 +1,5 @@
 /* eslint no-console:0 */
-define(function() {
+define(['mathlive/core/mathAtom'], function(MathAtom) {
 
 function findEndOfMath(delimiter, text, startIndex) {
     // Adapted from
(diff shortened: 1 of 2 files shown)


NSoiffer:
Some other file I'm in the middle of working on is preventing a build. Still learning the "gotcha"s of git and pull requests...  Will update and try again

PR #35, opened 2017-11-06
placeholder not showing up

### LINKED ISSUE #33: #0 as part of insert doesn't become a placeholder
## Description
Here is the documentation that describes how `#0` should work:
> .insert(content, options) insert the specified content at the current insertion point. With options it is possible to specify the insertion mode, as well as what will be selected after the insertion. If the content contains a #? a placeholder will be indicated in its stead. The #0 sequence will be replaced by the item currently selected (or a placeholder if nothing is selected)

In practice, doing an insert with `#0` does nothing with a few exceptions (fractions, square roots, ???). 

## Steps to Reproduce
Inline shortcuts are an easy way to see the problem.
1. open mathlive.io
2. type `alt-[`

What is inserted is \left[\right] -- no placeholder. If you type `ctrl-2`, you get a square root with a `radicand` placeholder.

Both shortcuts use `#0` (see editor-shortcuts.js).

The bug appears to be in  `Parser.prototype.scanImplicitGroup`

NSoiffer:
There seemed to be some confusion between using type '#' and type'placeholder in the code. I switched everything to be 'placeholder'. That's mostly cosmetic.

The real fix is that in `Parser.prototype.scanToken`, there was one `if` testing for '#' (before the change above) and another testing for 'placeholder'. The fix was to move the 'placeholder' test into the '#' test so that when there was not something to replace, a placeholder was inserted into the result.

Note: the previous code in that case (which probably was never true) used
the value of the placeholder token (typically a '0') for the placeholder.
That's not great visually, so I switch to using 'U+2753 = BLACK QUESTION
MARK ORNAMENT' which was used elsewhere.

Now `alt-[` produces:  
image

Fixes #33.

### DIFF
--- a/src/core/lexer.js
+++ b/src/core/lexer.js
@@ -223,7 +223,7 @@ Lexer.prototype.makeToken = function() {
                 }
             }
             if (isParam) {
-                result = new Token('#');
+                result = new Token('placeholder');
                 next = this.get();
                 if (next >= '0' && next <= '9') {
                     result.value = parseInt(next);
--- a/src/core/parser.js
+++ b/src/core/parser.js
@@ -1045,7 +1045,7 @@ Parser.prototype.scanArg = function(parseMode) {
 
     // If this is a param token, substitute it with the
     // (optional) argument passed to the parser
-    if (this.hasToken('#')) {
+    if (this.hasToken('placeholder')) {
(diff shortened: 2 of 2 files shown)


arnog:
I've checked in a different fix for this issue.

PR #68, opened 2018-04-13
Fixed mathml not being in code tags.

mildlyincompetent:
`mathml` was listed as \`mathml\' and as a result did not appear in `code` tags, unlike the rest of the optional keys listed in the MASTON docs.


arnog:
Thanks for noticing this and for fixing it! The .html version of the document is generated from `tutorials/MASTON.md`, so I fixed the problem in the `.md` file directly.  Don't hesitate to submit changes to the `.md` file directly if you see anything else off.

Issue #102, opened 2018-08-10
Feature request: \href, \class, \cssId, \style commands


arnog:
Implement these commands as per http://docs.mathjax.org/en/latest/tex.html#html

arnog:
The commands `\class` and `\cssId` are now supported.

arnog:
There ahas been no request for either `\href` or `\style`, so closing.

PR #105, opened 2018-08-21
Adding basic support for \class commands

nickgravelyn:
I've got what I think is a basic working version of `\class` which is only part of #102. Even if you don't need this PR hopefully others looking for class support before you fix it officially can use this PR to see how to patch it themselves.


nickgravelyn:
Actually it seems this causes issues with the typography layouts so while it works for some basic cases it's definitely not good enough for the general case.

Issue #69, opened 2018-04-13
Convert from AMD to JS modules


arnog:
MathLive currently uses requirejs to implment AMD-style modules. Native JavaScript modules are now supported by the major browsers and Node.

We should:
- convert all the import/export statements in the JS files to be JS modules
- modify the package.json file and webpack/babel configuration accordingly

jboxman:
Hi. I think I can look at this.

jboxman:
It doesn't look like I can close this.

PR #122, opened 2018-10-26
Use native modules

jboxman:
- Initial pass using amd-to-as6
- Fix paths
- Fix test/test.js


jboxman:
This is an initial attempt at this. I haven't looked at what happens when `webpack` is run, but the unit tests pass locally.

Looks like I need to tackle `eslint` errors and update the PR.

arnog:
Awesome! Thanks for working on this. 
You will have some es-lint errors about modules being defined but never used. Those modules are in fact used, but they are not referenced, and so this is not an error.

For example, it is necessary to include the OutputLatex module. This module modifies, e.g., MathAtom to add to it some additional functions. However, the OutputLatex module itself is not referenced. Therefore, es-lint need to silenced about those "errors".

jboxman:
Hi! Let me know if there's anything else I can do with this. Thanks.

arnog:
LGTM. If you're done with it, I'll merge it into master. Feel free to close issue #69 with this PR if you'd like.

jboxman:
Cool. Merge away. Thanks!

PR #168, opened 2019-04-09
Reducing node required version to LTS

nickgravelyn:
A recent update to the package made it fail to install because the engine for Node became specified as the latest Node major version. This PR just changes that requirement to allow the 10.x LTS versions as that's what we are using and can't see anything in Mathlive that would require the latest version of Node. 

We were maintaining an entire fork just for this change but it makes it much harder for us to take updates, so it'd be great if the main package could reduce the required Node version a little bit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
   },
   "engines": {
     "npm": ">=6.7",
-    "node": ">=11.9"
+    "node": ">=10.0"
   },
   "browserslist": [
     "> 1%",


arnog:
Thanks for this PR. No real requirement to use Node beyond LTS, so happy to change to this.

nickgravelyn:
Thanks! 

Would you mind publishing a new patch version (0.27.1) with this change so that we can move back to an NPM version of Mathlive? If not, we'll just wait until your next release and then switch back.

arnog:
Done.

PR #178, opened 2019-04-24
Readaloud

NSoiffer:
Merge speak and read aloud code and make the shortcuts similar. There are some oddities. E.g, if you have `a+b` and the cursor is after the `a`, you hear (at least from SRE), "positive b" instead of "plus b". Perhaps the MathML that gets generated should tag the `+` with `form='infix'` although I don't know if SRE bothers to check for that. Still, SRE can't do anything smart if that's not done.

TBD: speak from start/to end. That appears to be much harder. E.g, speak to end when somewhere in a denominator -- what should be said about being in a fraction? about the numerator? It needs special speech code.


src/editor/editor-mathfield.js
```diff
@@ -3264,132 +3264,126 @@ MathField.prototype.$setConfig = function(conf) {
...
-    this.config.handleSpeak(text, this.config);
-}
-
 /**
- * @method MathField#speakSelection_
+ * 
+ * Speak some part of the expression, either with or without synchronized highlighting.
+ * 
+ * @param {string} amount (all, selection, left, right, group, parent)
+ * @param {boolean} withHighlighting - If true, synchronized highlighting of speech will happen (if possible)
```
arnog: Use a named param rather than a boolean, so that the call can be `speak_("all", {withHighlighting: true})` rather than `speak("all", true)` and provide a default value for `withHighlighting`.

src/editor/editor-mathfield.js
```diff
@@ -3264,132 +3264,126 @@ MathField.prototype.$setConfig = function(conf) {
...
+ * 
+ * @method MathField#speak_
  */
-MathField.prototype.speakSelection_ = function() {
-    let text = "Nothing selected.";
-    if (!this.mathlist.isCollapsed()) {
-        text = MathAtom.toSpeakableText(this.mathlist.getSelectedAtoms(), this.config)
+MathField.prototype.speak_ = function(amount, withHighlighting) {
+    const getAtoms = function(mathField, amount) {
```
arnog:
```suggestion
    function getAtoms (mathField, amount) {
```

src/editor/editor-mathfield.js
```diff
@@ -3264,132 +3264,126 @@ MathField.prototype.$setConfig = function(conf) {
...
+                break;
+            case 'parent': {
+                const parent = mathField.mathlist.parent();
+                if (parent && parent.type !== 'root') {
+                    result = mathField.mathlist.parent();
+                }
+                break;
+            }
+            default:
+                console.log('unknown atom type "' + mathField.type + '"');
```
arnog: That doesn't seem to be the right error message...

src/editor/editor-mathfield.js
```diff
@@ -3264,132 +3264,126 @@ MathField.prototype.$setConfig = function(conf) {
...
-    if (!this.mathlist.isCollapsed()) {
-        window.mathlive.readAloudMathField = this;
-        this._render({forHighlighting: true});
-        const options = this.config;
-        options.textToSpeechMarkup = (window.sre && options.textToSpeechRules === 'sre') ? 'ssml_step' : 'ssml';
-        const text = MathAtom.toSpeakableText(this.mathlist.getSelectedAtoms(), options)
-        this._speakWithSynchronizedHighlighting(text);
-    } else {
-        this._speak("Nothing selected.");
+    const getFailedSpeech = function (amount) {
```
arnog:
```suggestion
    function getFailedSpeech (amount) {
```

src/editor/editor-mathfield.js
```diff
@@ -3264,132 +3264,126 @@ MathField.prototype.$setConfig = function(conf) {
...
+    if (withHighlighting) {
+        options.textToSpeechMarkup = (window.sre && options.textToSpeechRules === 'sre') ? 'ssml_step' : 'ssml';
+    }
+    const text = MathAtom.toSpeakableText(atoms, options)
+
+    if (withHighlighting) {
+        window.mathlive.readAloudMathField = this;
+        this._render({forHighlighting: true});
+        if (options.handleReadAloud) {
+            this.config.handleReadAloud(this.field, text, this.config);
```
arnog: the use of `options` vs. `this.config` is confusing.

PR #205, opened 2019-05-21
fix typo: &@x03c0; => &#x03c0;

johanneswilm:


arnog:
Great catch. Thank you!

PR #217, opened 2019-05-31
Fix incorrect loop-condition

Khazuar:
This fixes an error I got when pasting `\frac{n\cdot \left(n+1\right)}{2}` into a mathfield with `removeExtraneousParentheses` active by default.


arnog:
Oops. Good find and thanks for the fix.

PR #237, opened 2019-08-01
Add parameter type to MathFieldCallback

Khazuar:
This makes `MathFieldCallback`s much more useful.


arnog:
Oopps.. Yes, of course. Thanks for the PR!

Issue #238, opened 2019-08-02
Typing error this.focus() instead of this.$focus()


plfort:
**Describe the bug**
Small typing error in https://github.com/arnog/mathlive/blob/7a66b0c8eee002393d5311b5eddfc07d86f9ca9e/src/editor/editor-mathfield.js#L1055
Any button pressed on the virtual keyboard involving performWithFeedback_ function doesn't work.

**To Reproduce**
Got to https://mathlive.io/deploy/examples/customization/index.html, open JS console, open the virtual keyboard, click on "<" or ">".
In the JS console:
Uncaught TypeError: this.focus is not a function

**Environment**
Any

plfort:
Just realized that there are 2 buttons "<" and ">", I meaned the buttons involving performWithFeedback_ :
moveToPreviousChar
moveToNextChar
moveToNextPlaceholder
etc.

arnog:
Fixed with PR#239

PR #239, opened 2019-08-02
Fix performWithFeedback_ focus

plfort:
Fix issue #238


arnog:
Good catch. Thanks for the report and the PR!

PR #433, opened 2020-04-26
collect z-index of keyboard in one variable

spirobel:
Discourse also works with z-index values. I want to add mathlive to a modal so that people can use it to insert math equations into their forum posts. [https://meta.discourse.org/t/implementing-a-new-z-index-system-with-sass/78236](https://meta.discourse.org/t/implementing-a-new-z-index-system-with-sass/78236) So I had to change the z-index value to something higher than the modal, otherwise it wouldnt show up. This commit makes it possible to change the z-index of the keyboard in one place.
Thank you very much for creating mathlive!
Kind Regards,
Spirobel


arnog:
Thanks for this.

PR #446, opened 2020-05-10
Handle lineendings better in different environments

Khazuar:
After reading https://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/ and hopefully understanding that git is translating my local _text_ files from CRLF to LF in the repo and vice versa, prettier (which runs as part of the local build process) should accept the native line-ending format for my local system, too.
I am not 100% sure about the other files that we automatically included by the git-hook into the commit. Is that supposed to be happening like this? @arnog


arnog:
I hadn't paid attention to the Windows side of things in a little while, so yeah, sorry. Thanks for that PR. I think it's correct and the CR/LF should be taken care of by Git without requiring your editor to switch its settings.

The /dist files are there because of a change in the build system I made that automate the updating of the dist folder on each commit. It seemed like a good idea at the time, but now I'm having second thoughts as it adds lots of noise to every single commit.

OK, I've manually merged this change and a few other improvements to get the build environment working better on Windows.

arnog:
OK, manual merged done. Tested on Windows and things seem to work. Fingers crossed.

Issue #208, opened 2019-05-22
Unable to focus on a mathfield div on mobile


manuelsabu:
**Describe the bug**
If the MathField is set to "onfocus", the keyboard does not show up when the user clicks the related div. The "manual" option works fine to open the keyboard, however it's not possible to move the cursor to a specific position in the equation using touch.

**To Reproduce**
Steps to reproduce the onfocus behavior:
1. Create a mathField with "virtualKeyboardMode": "onfocus"
2. Use mobile Safari or Chrome to click on the mathfield div
3. Keyboard doesn't pop up

Steps to reproduce the lack of focus with the cursor
1. Create a mathField with "virtualKeyboardMode": "manual"
2. Use mobile Safari or Chrome and toggle the virtual keyboard
3. Type something using the keyboard
4. Touch anywhere in the displayed text to attempt to move the cursor
5. Unable to move the cursor using touch

**Expected behavior**
The MathField should maintain focus on mobile devices so that the "onfocus" functionality works, as well as moving the cursor works.

**Environment**
- Device: [iPhone]
- OS: [e.g. iOS 12.2]
 - Version [e.g. Chrome 74, Safari 12.2]

leerobert:
@arnog any ideas on how we can fix an environment specific issue like this? I'd like to help

arnog:
I believe this is fixed now. Let me know if this is still a problem for you.

camjosen:
I can still repro:

Mathlive version: 0.35.1
Device: [iPhone 6s]
OS: [e.g. iOS 12.4.1]
Browser [e.g. chrome, safari]
Version [e.g. Chrome 81, Safari 12.2]

arnog:
To clarify: this is fixed in the master branch. The version on npm (35.1) is lagging behind.

PR #449, opened 2020-05-14
Fix focus bug on iOS 12 and Firefox/Android

beneater:
(and maybe others?)

Test plan:
- Visit /examples/basic/index.esm.html on an iPhone 6
- Tap into mathfield; it gets focus
- Tap around mathfield to move cursor

### DIFF
--- a/src/editor/mathfield-pointer-input.ts
+++ b/src/editor/mathfield-pointer-input.ts
@@ -20,7 +20,7 @@ export function onPointerDown(mathfield: MathfieldPrivate, evt) {
     let trackingWords = false;
     let dirty = false;
     // If a mouse button other than the main one was pressed, return
-    if (evt.buttons !== 1) {
+    if (evt.buttons !== 1 && evt.buttons !== 0) {
         return;
     }
     let scrollLeft = false;


beneater:
Hi there! We’ve been using MathLive for the past few weeks in a project we’ve been developing for the [Lockdown Math live video series](https://itempool.com/c/3b1b). MathLive has an impressive amount of functionality!

We had thousands of users banging away on it and received a half-dozen bug reports, all related to not being able to focus the mathfield. The issue seemed to only affect a few iOS safari users and some Firefox users on Android.

I was able to reproduce the issues and the behavior is precisely as described in #208, so maybe that issue hadn’t been completely resolved.

In any event, I could reliably reproduce the issues on Safari on iOS 12.4, which is the latest version of iOS available for a number of devices that are still in use, e.g iPhone 6/7.

The issue is that it appears that in iOS 12.4 Safari and Firefox on Android, the `touchstart` event is sent with `event.buttons = 0` which for a mouse event would normally be an invalid button. There’s some logic in `mathfield-pointer-input.ts` that ignores non-primary-button presses that was also ignoring this case. Consequently the fix appears to be pretty straightforward.

arnog:
Awesome. Thanks for letting me know and for providing a fix. Yeah, it seems like a reasonable fix. 

As an aside, big fan of your work, and love the Lockdown Math live video series (which I didn't know about, but looks really cool). Let me know if you get any other feedback from users that could help improve MathLive.

arnog:
I've published 0.50.8 which incorporates this fix as a GitHub Release and on npm.

beneater:
Awesome, thanks @arnog! Appreciate the release on npm too! We'll let you know as we get more feedback, but so far we've been pretty happy with MathLive.

PR #521, opened 2020-06-27
Fix broken typescript definitions path

OrkhanAlikhanov:
After moving typescript definitions to public folder, typescript was unable to find type definitions. This fixes that.

@arnog If we had an e2e test for this, it would've been detected earlier.

Thanks for your efforts 👍


arnog:
Great, thank you!

PR #594, opened 2020-08-23
fix: Add more ways to move out of mathfield #591

### LINKED ISSUE #591: Add more ways to move out of mathfield
## Prerequisites

-   [x] New issue

## Description

It is possible to leave the math field using
- `Right` and `left` arrow keys
- `Tab`

However, a few other editors that I'm used to also make it possible to "submit" an expression. This means, when the user hits `Enter` and isn't in command mode, the mathfield loses focus. 
Furthermore, it would be great if the `up` and `down` arrow keys could also be used to leave a math field. This is quite useful when you have multiple math fields in a row. [An example of the behavior I'm looking for.](https://www.desmos.com/calculator)

### Actual behavior

The mathfield keeps focus

### Expected Behavior

It should be possible to leave the mathfield using those keys as well
- Enter
- Up or down

### Notes

I sadly don't think that this can simply be implemented in userland by using the shortcuts feature nor with `onKeystroke`.

 Lastly, I'm more than willing to open a PR that implements this. However, I'd need a bit of guidance regarding where to start.

stefnotch:
Fixes #591

However, this doesn't add wrapping around for `upward`s and `downward`s movement.

### DIFF
--- a/src/editor/model-selection.ts
+++ b/src/editor/model-selection.ts
@@ -635,6 +635,9 @@ export function up(
         }
     } else {
         model.announce('line');
+        if (!model.suppressChangeNotifications) {
+            model.hooks?.moveOut(model, 'upward');
+        }
     }
     return true;
 }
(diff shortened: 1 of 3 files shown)


stefnotch:
Fixed

arnog:
Looks good! Thanks for this contribution!

PR #595, opened 2020-08-24
fix: Don't throw an error when the first argument of \enclose is empty

stefnotch:
Previously, the first argument could be an empty array `[]`. Actually, maybe that's a bug with the argument parsing function?


stefnotch:
Good point, I forgot to consider that. Fixed.

arnog:
Great. Thanks for your contribution. I really appreciate it.

PR #612, opened 2020-09-03
Fix issue when typing "sqrt"

beneater:
Previously, typing "sqrt2" would insert "\sqrt[]{}2". With this change, typing "sqrt2" will insert "\sqrt[]{2}".


arnog (approved):
Nice find! Thanks for the fix.

arnog (approved):

PR #614, opened 2020-09-05
Improve getConfig typings

stefnotch:
This improves the public `getConfig()` type for the single key case. 
e.g. `mathfield.value.getConfig("keybindings")` will now return a `Keybinding[]`.

@ arnog 
If you want, I can also change the internal `getConfig` methods to use these typings. Also, if it would be useful, I can try to create a fully typesafe version of `getConfig(keys: string[]): MathfieldConfig;` Regarding the types, why does the last overload exist? Furthermore, why is `getConfig` the only method that doesn't have a `$`?

Furthermore, while I'm at it, if there are any other bits and bobs that could use some extra type safety, I'd be more than happy to take a look at them!

### DIFF
--- a/src/public/mathfield.ts
+++ b/src/public/mathfield.ts
@@ -82,7 +82,7 @@ export type InsertOptions = {
 export interface Mathfield {
     mode: ParseMode;
 
-    getConfig(keys: keyof MathfieldConfig): any;
+    getConfig<T extends keyof MathfieldConfig>(key: T): MathfieldConfig[T];
     getConfig(keys: string[]): MathfieldConfig;
     getConfig(keys: keyof MathfieldConfig | string[]): any | MathfieldConfig;
 


arnog:
This is awesome, thanks! Much better, indeed.

Regarding your questions:
> Also, if it would be useful, I can try to create a fully typesafe version of getConfig(keys: string[]): MathfieldConfig; 

Sure, that would be great. No idea how to go about it, but there might be a way...

>Regarding the types, why does the last overload exist?

Great question. The last overload is only needed for the implementation (it's the union of the other overloads). However, for the public interface, I don't think it's needed, and it's probably confusing to have it here.

> Furthermore, why is getConfig the only method that doesn't have a $?

Because now that there are cleanly defined interfaces, I plan on removing all the "$" prefix (they were needed before to distinguish between public/private). `getConfig` was added after the public interface were added, and thus didn't inherit the "$" suffix.

> Furthermore, while I'm at it, if there are any other bits and bobs that could use some extra type safety, I'd be more than happy to take a look at them!

Hmmm... there are probably a few other places. One that comes to mind is `Selector`, and in particular when there is a `Selector | any[]` type. The `any[]` is meant to describe a tuple, with the first element a `Selector`, and the rest some various types. With the variadic tuple types in TypeScript 4, this should now be able to be represented with a more precise type.

PR #627, opened 2020-09-06
Improve getConfig and selector types

stefnotch:
This PR slightly improves some types.

I chose to use `[SelectorPrivate, ...any[]]` for now, because the August 2020 version of Visual Studio Code, which comes with Typescript 4.0, hasn't been released yet. (The expected release date is September 9)

By the way, are such minor improvements to the types `fix`es or `feat`ures?

### DIFF
--- a/src/public/mathfield.ts
+++ b/src/public/mathfield.ts
@@ -88,7 +88,7 @@ export interface Mathfield {
     getConfig<K extends keyof MathfieldConfig>(key: K): MathfieldConfig[K];
     getConfig(): MathfieldConfig;
 
-    $setConfig(config: MathfieldConfig): void;
+    $setConfig(config: Partial<MathfieldConfig>): void;
 
     /**
      * Reverts this mathfield to its original content.
--- a/src/public/mathfield.ts
+++ b/src/public/mathfield.ts
@@ -82,9 +82,11 @@ export type InsertOptions = {
 export interface Mathfield {
     mode: ParseMode;
 
-    getConfig<T extends keyof MathfieldConfig>(key: T): MathfieldConfig[T];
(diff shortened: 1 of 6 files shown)


src/public/mathfield.ts
```diff
@@ -82,11 +82,13 @@ export type InsertOptions = {
...
-    getConfig(keys: string[]): MathfieldConfig;
-    getConfig(keys: keyof MathfieldConfig | string[]): any | MathfieldConfig;
+    getConfig<K extends keyof MathfieldConfig>(
+        keys: K[]
+    ): Pick<MathfieldConfig, K>;
+    getConfig<K extends keyof MathfieldConfig>(key: K): MathfieldConfig[K];
+    getConfig(): MathfieldConfig;
 
-    $setConfig(config: MathfieldConfig): void;
+    $setConfig(config: Partial<MathfieldConfig>): void;
```
stefnotch, line 91: This doesn't actually have an effect, because everything in the `MathfieldConfig` is optional, but at least it makes it a bit more explicit that you can pass in a partial object.

arnog: Yeah, I like that. I'm wondering if `MathfieldConfig` shouldn't have optional fields, but instead just use `Partial<>` when appropriate. Seems like a clearer convention...

arnog (approved):
All those changes look good to me. Thanks!

arnog:
I would consider these `fix`es. I try to reserve `feat`ures for things that you could name (" the `foo` feature is available in this version") or write a blog post about.

Thanks for these improvements. Much better!

src/public/mathfield.ts
stefnotch: That does sound like a reasonable idea. 👍

PR #636, opened 2020-09-13
Added getCaretPosition() and setCaretPosition()

### LINKED ISSUE #101: Feature Request: API to read/set the caret location based on screen coordinates
Make _getCaretClientPosition() public and add a setCaretAtPosition() method.

stefnotch:
I added public `getCaretPosition()` and `setCaretPosition()` methods. This fixes #101 

However, I do have some open questions that I'd like to address:

**getCaretPosition**
- It can return `null`. Is that fine? Or should it return `undefined`? 

**setCaretPosition**:
- Should it call `model.announce('move', ...)`?
- Should it focus the mathfield when calling it?
- Is the boolean return value fine or should it pick the nearest point when it is out of bounds?

### DIFF
--- a/src/editor/keybindings-definitions.ts
+++ b/src/editor/keybindings-definitions.ts
@@ -69,8 +69,8 @@ export const DEFAULT_KEYBINDINGS: Keybinding[] = [
         command: 'moveToPreviousPlaceholder',
     },
 
-    { key: '[Escape]', ifMode: 'math', command: ['switch-mode', 'command'] },
-    { key: '\\', ifMode: 'math', command: ['switch-mode', 'command'] },
+    { key: '[Escape]', ifMode: 'math', command: ['switchMode', 'command'] },
+    { key: '\\', ifMode: 'math', command: ['switchMode', 'command'] },
 
     {
         key: 'alt+[Equal]',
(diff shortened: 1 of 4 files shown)


arnog:
- Returning `null` when there's no valid caret position is fine. `undefined` in general indicate a value that has not been initialized, while `null` is an object indicating "no value"
- Yes, `setCaretPosition` should call `model.announce` for accessibility support
- I would not focus the mathfield when `setCaretPosition` is called. There might be use cases where that would be undesirable. On the other hand, if focusing is necessary, calling `focus()` is trivial.
- I think the boolean return value is fine. I think the intent of `setCaretPosition()` is to be the dual of `getCaretPosition()`

Thanks for the work on this!

stefnotch:
Great! I added the call to `model.announce`, now it should be ready for reviewing and/or merging.

src/editor/keybindings-definitions.ts
```diff
@@ -69,8 +69,8 @@ export const DEFAULT_KEYBINDINGS: Keybinding[] = [
         command: 'moveToPreviousPlaceholder',
     },
 
-    { key: '[Escape]', ifMode: 'math', command: ['switch-mode', 'command'] },
-    { key: '\\', ifMode: 'math', command: ['switch-mode', 'command'] },
+    { key: '[Escape]', ifMode: 'math', command: ['switchMode', 'command'] },
```
arnog, line 72: Is there a reason for this change? Note that both "switchMode" and "switch-mode" are valid perform selectors. Or is this because of the stricter Typescript type checking for selectors?

arnog (commented):
Looks good. Have a question of the `switch-mode` change.

src/editor/keybindings-definitions.ts
stefnotch: It is because of the stricter Typescript type checking. Typescript sadly can't do fancy string transforms.

PR #638, opened 2020-09-14
Fix non canonical text characters

### LINKED ISSUE #599: German Umlaute in `\text{}` get removed
## Description

German umlaute in pasted text-areas like `\text{für}` get removed. Inside `\operatorname`s they remain in latex, but don't show either.

image

### Expected Behavior

This used to work fine in the 0.35.x-versions.

## Steps to Reproduce

Paste `\text{für}` into the mathfield on mathlive.io.

### Environment

**Operating System**
Windows 10

**Browser**
Chrome

stefnotch:
Fixes #599 
(arnog, thank you very much for the guidance)

There are 3 further functions where `TEXT_SYMBOLS` is being used. I'm not sure if I should add anything to them.
- `commandAllowed`: It isn't being called anywhere
- `getValue`: It is only being called with 'math'
- `emit`: What is this function used for?

### DIFF
--- a/src/core/modes-text.ts
+++ b/src/core/modes-text.ts
@@ -293,7 +293,7 @@ function parse(
             let atoms: Atom[];
             [atoms, tokens] = options.parse('text', tokens, options);
             result = [...result, ...atoms];
-        } else if (token.length === 1) {
+        } else if ([...token].length === 1) {
             const info = getInfo(token, 'text', options.macros);
             if (!info) {
                 error({ code: 'unexpected-token' });
--- a/src/core/modes-text.ts
+++ b/src/core/modes-text.ts
@@ -293,7 +293,20 @@ function parse(
             let atoms: Atom[];
             [atoms, tokens] = options.parse('text', tokens, options);
             result = [...result, ...atoms];
-        } else if ([...token].length === 1) {
+        } else if (token === '<$>' || token === '<$$>') {
+            // Mode-shift
+            const subtokens = tokens.slice(
+                0,
+                tokens.findIndex((x: Token) => x === token)
+            );
+            tokens = tokens.slice(subtokens.length + 1);
+            const [atoms] = options.parse('math', subtokens, options);
+            result = [...result, ...atoms];
+        } else if (token === '<{>' || token === '<}>') {
+            // Spurious braces are ignored by TeX in text mode
+            // In text mode, braces are sometimes used to separate adjacent
+            // commands without inserting a space, e.g. "\backlash{}command"
+        } else {
             const info = getInfo(token, 'text', options.macros);
             if (!info) {
                 error({ code: 'unexpected-token' });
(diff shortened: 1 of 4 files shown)


src/core/modes-text.ts
```diff
@@ -293,7 +293,7 @@ function parse(
             let atoms: Atom[];
             [atoms, tokens] = options.parse('text', tokens, options);
             result = [...result, ...atoms];
-        } else if (token.length === 1) {
+        } else if ([...token].length === 1) {
```
stefnotch, line 296: This is for correctly handling characters like 💩 that take up two UTF-16 characters. Is that fine?
(See https://mathiasbynens.be/notes/javascript-unicode )

arnog: OK, it took me a while to figure out (a) what your fix was doing and (b) what was the underlying problem, but I got it now. It's a good idea, but it's only a partial fix, and it has some performance implications (converting the string to an array of codepoints with the spread operator for each token).

It's only a partial fix because of a subtlety of Unicode encoding where some characters can actually be more than two codepoints, and wouldn't be correctly handled (or are not guaranteed to be handled) by the Javascript spread operator. For example the David Bowie 👨🏻‍🎤 emoji is made up of U+1F468 (man) + U+200D (zero-width-joiner) + U+1F3A4 (microphone). There is already code in Mathlive that handles this (see `graphemesplitter`), so that this combination will be considered a single token.

It is indeed incorrect to check for the length here. However, the better fix is simply to move the code that's in the `if (token.length ===1)` branch into the `else {}` branch (and just remove the error call). If the token is *not* one of the special ones, we should just handle it, without checking the length of the token first.

arnog:
`commandAllowed`: yeah, good catch. There used to be more restrictions regarding what commands could be used in the different modes, but we're more liberal now. This entire function can be removed.

`getValue`: yes, good point. That could probably be simplified (i.e. remove the 'mode' argument)

`emit` is used to "emit" the formula into Latex. This code should convert the unicode sequence into an appropriate Latex, either one of the "well-known" (i.e. registered in `TEXT_SYMBOLS`) command, or a generic command like (`^^^^` or `\char`).

arnog (approved):
Looks good!

stefnotch:
Ok, I addressed the comments except that I left the emit function in `definitions-utils.ts` unchanged. (I realize that I should have specified which emit function I was referring to originally) https://github.com/arnog/mathlive/blob/be533c9ea6fef282a34f70677fa61f979b3bef5e/src/core/definitions-utils.ts#L762
I tested it out and am still not entirely sure what exactly this function does and when it gets called.

src/core/modes-text.ts
stefnotch: Oh, thank you! I learned something new about Unicode, that's great.

arnog:
The `emit` function in `definition-utils.ts` gets called to generate the Latex for a formula, for example as a result of a call to `mf.$text('latex')`. It should be modified so that if it encounters a symbol which is not in the `TEXT_SYMBOLS` or `MATH_SYMBOLS` table it generates some suitable Latex (either a `\char` command or a sequence of `^^`)

stefnotch:
Ok, I think I fixed the emit function.

arnog:
Looks good, thanks, and sorry for the delay, I missed that notification.

PR #639, opened 2020-09-16
fix: Handle backspace with regards to inline shortcuts

### LINKED ISSUE #198: Typing slowly yields different results
**Describe the bug**
Typing slowly yields different results
Making corrections too.

**To Reproduce**
1. Go to mathlive.io
2. Type sum or \sum but wait a couple of seconds after the s
3. You do not get a sum sign

image

Alternatively:
1. Go to mathlve.io
2. Type sun or \sun (yes, a mistake) and correct the n into an m
3. You do not get a sum sign

**Expected behavior**
Typing pace or corrections should not affect result

**Environment**
- Device: PC
- OS: Windows
 - Browser Chrome
 - Version v0.29.0

stefnotch:
Fixes #198 

This is a rather simplistic way of checking if backspace was pressed and then handling it.

It doesn't really handle the case where you type `su` on the normal keyboard, then hit `[Backspace]` on the virtual keyboard and then type `m` on the normal keyboard. In fact, a number of cases where the normal and virtual keyboard are mixed don't work as expected. Furthermore, the virtual keyboard doesn't ever modify the keystroke buffer, is that intentional?

Also, I noticed that mathlive doesn't handle `su` `Ctrl+V` `m` correctly either.

### DIFF
--- a/src/editor/commands.ts
+++ b/src/editor/commands.ts
@@ -76,7 +76,9 @@ export function perform(
     ) as SelectorPrivate;
     if (COMMANDS[selector]?.target === 'model') {
         if (/^(delete|transpose|add)/.test(selector)) {
-            mathfield.resetKeystrokeBuffer();
+            if (selector !== 'deletePreviousChar') {
+                mathfield.resetKeystrokeBuffer();
+            }
         }
         if (
             /^(delete|transpose|add)/.test(selector) &&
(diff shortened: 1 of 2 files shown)


arnog (approved):
looks good to me

arnog:
I think that's a good fix. Certainly an improvement over the current behavior.

The idea of the inline shortcuts is that they somewhat depend on timing, and are therefore most appropriate with a hardware keyboard. The virtual keyboard having additional virtual keys, the idea was also that it would be less necessary/useful to also have inline shortcuts (since those "shortcuts" can be actual virtual keys). Similarly, handling the clipboard commands was felt to be a very niche use case. That said, it would not be completely unreasonable to handle those, but that would probably best be viewed as a separate feature enhancement.

PR #662, opened 2020-10-03
editor/parse-math-string: pasting/writing `sqrt` should behave the same

AceGentile:
in math editor writing a `sqrt` string is translated to a
`\\sqrt{\\placeholder{}}` expression. Pasting the same `sqrt` string
into the editor is translated into a `\\sqrt{}` expression.

This happens because `parseMathExpression` does not provide
a fallback value when sqrt / sqrt[3] values are empty


arnog:
Looks good to me. Thanks for finding this and for the fix.

Issue #698, opened 2020-11-06
Feature request: disable toolbar on virtual keyboard


alexprey:
## Description

Whan about to add new option to make it possible to disable toolbar on virtual keyboard (copy, undo, redo)?
![image](https://user-images.githubusercontent.com/7888563/98396507-dbbcb180-206e-11eb-8a5a-25be15255ab9.png)

Sometimes it can be handfully, for example I'm want to extract this functions to external place and handle changes between multiple math fields at the same time. As a workaround I can use styling to hide this block on virtual keyboard, but this is not a clear solution.

alexprey:
@arnog if you approve this feature I can introduce that on the next week

arnog:
Yes, I think that makes sense. It would probably make sense as a `virtualKeyboardToolbar` property of `VirtualKeyboardOptions` with values of `none` and `default`. This gives the option of adding other values later, including some custom toolbars.

alexprey:
Done, I'm add a new option `toolbar` with two possible values - `none` and `default` as we discuss above

arnog:
Great, thanks for the PR! Closing now.

alexprey:
Thanks ^^

ShaMan123:
Drop a thought. Do you guys think it makes sense to add a close button, hiding the keyboard?

alexprey:
@ShaMan123 for me it looks as a good idea, but need to think about behavior of this button together with keyboard mode setting. For manual mode it is ok, but what about when onfocus mode choosen, should we hide it?

ShaMan123:
Good point. 
Perhaps it should be hidden if no keyboard toggle glyph is present because if it is present then after dismissing the keyboard one could show it again by pressing the glyph.

arnog:
It would seem a bit confusing to have this button and the virtual keyboard toggle button. Wouldn't they be doing the same thing?

ShaMan123:
I guess so.
Strike out the motion.

PR #705, opened 2020-11-11
Feature/#698 keyboard toolbar config

alexprey:
Implement the feature request by #698


arnog (approved):
Looks good. Thanks for the PR.
The only change I would make is to rename the `toolbar` option to `virtualKeyboardToolbar` to be more explicit

arnog (approved):
Removed unnecessary import of `VirtualKeyboardToolbarOptions` which caused the build to fail and renamed `toolbar` to `virtualKeyboardToolbar`

Issue #727, opened 2020-11-25
The alternate keys panel renders the wrong layout for the "logic" (∀) key


Kydyralin:
## Description

The alternate keys panel renders the wrong layout for the "logic" (∀) key.

## Steps to Reproduce

[Provide steps that are specific and repeatable, if possible]

1. Go to [https://mathlive.io](https://mathlive.io) page
2. Click the 'keyboard' icon on the formula input
3. Click and hold the "logic" (∀) key till the alternate keys panel will popup

### Actual Behavior
[What happens when you follow the steps above]
<img width="701" alt="Screen Shot 2020-11-26 at 2 31 11 AM" src="https://user-images.githubusercontent.com/22542620/100285068-088b2700-2f92-11eb-8a7e-b02640e55063.png">

[If there are any error messages, include the exact text shown,
or upload a screenshot. Some error messages may be displayed in the Javascript console.]

### Expected Behavior

[What did you expect to happen insteead]
<img width="700" alt="Screen Shot 2020-11-26 at 2 47 17 AM" src="https://user-images.githubusercontent.com/22542620/100284928-cd88f380-2f91-11eb-9a09-0ed5b65febbb.png">

[Is this a regression: did it use to work in a previous version?]

### Environment

**Operating System**
[macOS, Windows, iOS. Include the version if possible]
macOS, Windows

**Browser** [Safari, Chrome, IE, Firefox, etc...]
It's not browser-specific as it can be reproduced on Safari, Edge, Chrome browsers.

arnog:
Note: the viewport height has to be below a specific threshold for this issue to occur.

Kydyralin:
Hi, I've found a piece of code on **virtual-keyboard-commands.ts** file that sets the height for the panel, and considering the comments there it was designed with the assumption that in the panel would no more than 3 rows with alternate keys. Probably it is a regression bug. As a temporary fix, I can add the logic to handle up to 7 rows. What do you think about it? 
I know that it's not a generic solution but as far as there are no alternate keys panel with more than 28 elements it would work fine :)

arnog:
Ah... right, I remember now. There is a maximum to the number of keys on a pop-up panel. I think it's 28 (4 rows / 6 columns). But this many keys are only laid out correctly when using a "compact" style (keycaps are 50x50 instead of 70x70), but this only triggers via a CSS rule if the viewport is less than 412px tall. 

[...]

PR #728, opened 2020-11-27
Fix issue #727: The alternate keys panel renders the wrong layout for the "logic" (∀) key

Kydyralin:
Hopefully, I got your idea right in the discussion of the issue.

So here I've made a property evaluation for alternate keys based on variables, during the test no issues were found. Anyway if you have any concerns or suggestions to make it in another way, feel free to let me know.


arnog (approved):
Looks great. Thoughtful use of CSS custom properties (and thanks for fixing the typo in the build script). Thanks for the contribution!

Issue #737, opened 2020-12-01
Matrix braces is not supported in MathML: In conversion to MathML, unknown type : mopen


alexprey:
## Description

Usage of braces from virtual keyboard leads to errors of MathML converter, that says that `mopen` or `mclose` entries is not supported. But if I use physical keyboard with `[` the output is correct

![image](https://user-images.githubusercontent.com/7888563/100737958-f243f780-33e5-11eb-87f9-3d5e95527597.png)

## Steps to Reproduce

1. Use math-field element with MathML output
2. Place matrix braces from virtual keyboard

### Actual Behavior

Output is not genereted correctly, debug information in console:
```
In conversion to MathML, unknown type : mopen
```

### Expected Behavior

Correct MathML are generated w/o any errrors.
Matrix specification: https://www.w3.org/TR/MathML3/chapter3.html#presm.mtable

### Environment

**Operating System** Windows
**Browser** Opera

arnog:
Fixed now.

PR #738, opened 2020-12-01
fix the typo of ApplyFunction hexadecimal value

Kydyralin:
Hi,

I've noticed that there is a typo on the ApplyFunction hexadecimal symbol value so I decided to fix it. Not sure that it is related to the [issue #737](https://github.com/arnog/mathlive/issues/737) but seems that this fix makes the converter logic work a bit better (according to the console logging messages). If you have any suggestions please let me know and I'll try to apply them.

### DIFF
--- a/src/addons/math-ml.ts
+++ b/src/addons/math-ml.ts
@@ -29,13 +29,9 @@ const SPECIAL_OPERATORS = {
     '\\hat': '&#x005e;',
 };
 
-const HEXADECIMAL_SYMBOLS = {
-    applyFunction: '&#x2061;',
-};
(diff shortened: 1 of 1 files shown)


arnog (changes requested):
Thank you for catching this and submitting a fix.
Good idea to use constants for those values, but I would just define the constants directly, i.e. `const APPLY_FUNCTION = '&#x2061'`. The fact that they are encoded as hexadecimal or decimal is not very important.

Kydyralin:
Thanks for the quick response and review. Pushed a commit with the refactored constants.

arnog (approved):
Looks good. Thanks!

PR #753, opened 2020-12-07
Feature: Remote keyboard

alexprey:
Hi!

I'm implement a new feature that was required for our special cases and I hope it can be usufull for someone else. By this PR I'm also introduce small code refactoring around virtual-keyboard integration and mention few places for the next possible refactoring and cleanup.

# The Feature: Description

By our requirement we need to render mathfields in multiple iframes with small area for content and placing the virtual keyboard in that was is not really possible. I propose the idea to extract virtual-keyboard logic to separate control and render it on parent frame, allow it to manage multiple frames at same time. We implement it and it starts to work. I'm add examle page in this repository to check how it works.

remote-keyboard

# Design

Each `mathfield` element with `useProxyHost` option use the special host class to which mathfield element deligates all command execution. All commands related to virtual-keyboard sends to parent frame, all other commands handled internally in this frame.

On parent frame side the special instance of remote client manage all commands received from virtual-keyboard and route it to correct child frame that have focused mathfield element at this moment. And all virtual-keyboard commands handled internally by current frame.

# Technical notes

To done this feature I made refactoring around integration between virtual-keyboard and mathfield element. Actually before refactoring most of iteractions logic was implemented with commands, but some kind was implement by direct call of mathfield methods. That makes imposible to implement remote logic. I extract this logic to commands and replace direct calls to command execution with minimal risks.

Communication between frames was implemented with `postMessage(...)` method of parent and child frame windows. With `proxyHostTargetOrigin` and `targetOrigin` options it possible to control the `targetOrigin` parameter of `postMessage` on parent and child frame sides.

The new options for mathfield element was added: 
- `useProxyHost`
- `proxyHostTargetOrigin`


arnog:
Thanks for this pull request!
This is a rather large submission, but I understand what you want to do.
I think it could be simplified by dropping a few outdated requirements and with a bit of refactoring. I'm going to work on that and I'll submit you an updated PR.

alexprey:
Great! Thanks, please tell me what can I do to help you with completing/refactoring of this PR

arnog:
Thanks again for this pull request. That's a really awesome one. I think you did a fantastic job with it. 

I have pushed some changes to your branch. Please have a look at them and if you're happy with them, merge them with this PR and I'll be happy to merge it into the main branch. 

The main changes I made were some refactoring and simplification of the way the virtual keyboard and mathfield interact with each other. The architecture that was in place before didn't really make sense (that was my fault, the changes you made were reasonable), and you did a great job to work around it and make it work, but I figured that this was an opportunity to refactor the old architecture, which ended up making a lot of things simpler.

There are still a few optional things left to do, but nothing super important (except maybe more strict checks on `origin` in the message handlers)

src/public/mathfield-element.ts
```diff
@@ -337,7 +337,7 @@ export class MathfieldElement extends HTMLElement implements Mathfield {
             'virtual-keyboard-mode': 'string',
             'virtual-keyboard-theme': 'string',
             'virtual-keyboards': 'string',
-            'use-proxy-host': 'boolean',
+            'use-shared-virtual-keyboard': 'boolean',
```
alexprey, line 340: I think that target origin attribute also should be introduced there

alexprey (commented):
Thanks for you changes for my feature PR. Thats is a perfect idea to make virtual keyboard as an abstract interface and deligate all logic. Due I don't well know architecture and history of this code, so done this feature with not perfect solution. Your changes really make feature more easier and logically right, thanks you!

src/public/mathfield-element.ts
arnog: ah, yes, you're right

alexprey (commented):
Thanks!

alexprey:
I retest integration with new changes and let you know if all works fine and PR can be merged

alexprey:
I'm test refactoring and found few issues with internal keyboard implementation. I'm try to fix it and it looks like that it work now, please check my changes around `VirtualKeyboard` class implementation and part of this instance initialization in `MathFieldPrivate` constructor.

Also I'm implement origin validation logic and add new option `originValidator` to configure behaviour in both options sources. And I'm change the default values to that with privacy by default politics, set `targetOrigin` options to `window.origin` and `originValidator` option to `'same-origin'`.

**`originValidator` option:**

Specify behaviour for origin validation.
| Value | Description |
| ----- | ----------- |
| `'same-origin'` | The origin of received message must be the same of hosted window, instead exception will thow. |
| `'(origin: string) => boolean` | The callback to verify origin to be expected validation. When callback return `false` value, message will rejected and exception will throw. |
| `'none'` | No origin validation for post messages. |

alexprey:
@arnog I thinks that this PR are ready, if you agree with me, then you can merge it.

Thanks for colloboration and great library!

arnog (approved):
Excellent! Looks good to me.

Issue #726, opened 2020-11-25
'x' symbol pressed several times leads to strange behavior  


stylebc:
## Description

When I'm typing 'x' 2 times it is replaced with a cross symbol. 
![image](https://user-images.githubusercontent.com/7346159/100220756-d85b6e00-2f28-11eb-93a3-7534542ab659.png). 
Then repeat it 3 times, `x` symbol appears in unexpected places (or expected, not sure here). 
Then remove all by a backspace and enter the `x` symbol again. It will insert not `x` symbol but several crosses
gif attached 

## Steps to Reproduce

1. Press `x` on keyboard 6 times
2. remove symbols by backspace
3. press `x` again

### Actual Behavior
several crosses appeared
![chrome-capture (3)](https://user-images.githubusercontent.com/7346159/100222532-49038a00-2f2b-11eb-8ded-cf62f38cc073.gif)

### Expected Behavior
one `x` symbol should be inserted

### Environment

**Browser** Chrome

arnog:
By default, "xx" is an inline shortcut for the `\times` command (✖️). This inline-shortcut can be turned off using the [inlineShortcuts`](http://cortexjs.io/docs/mathlive/#(%22options%22%3Amodule).(InlineShortcutsOptions%3Atype)) option.
So, that behavior is expected.
However, there is something going on after backspace is pressed. Either a single "x" or a `\times` (✖️) character should be inserted, but a string of them can appear, which is not the correct behavior.

alexprey:
Not sure that this flow is correct. What I expect when typing 4 x symbols

| typings | actual | expected |
| - | - | - |
| `x` | `x` | `x |
| `xx` | `✖️` | `✖️` |
| `xxx` | `x✖️` | `✖️x` |
| `xxxx` | `✖️✖️` | `✖️✖️` |
```

This are actual for all shortcuts.

I'm try to looking into this issue and found that you use keystroke buffer that handled to rendering items. Why it done by this way?

PR #759, opened 2020-12-10
Fix issue #726: Wrong behaviour of keystroke buffer

alexprey:
Fix the [following issue #726](https://github.com/arnog/mathlive/issues/726)

# Cause 1
The first cause of wrong behaviour that is incorrect option validation in `resetKeystrokeBuffer` method.
It was a `if (this.options.inlineShortcutTimeout) {` and default value of `inlineShortcutTimeout` options is `0` that leads to `false` evaluation of condition and keystroke buffer is not reseted with `defer` option.

# Cause 2
The logic for resetting the keystroke buffer was strange and actually reseted everytime after fixing the first cause, so I change logic to reset buffer only when shortcut to insert was found. I'm not sure that new logic is correct, please mention if I not check some special cases for shortcuts

### DIFF
--- a/src/editor-mathfield/keyboard-input.ts
+++ b/src/editor-mathfield/keyboard-input.ts
@@ -166,12 +166,8 @@ export function onKeystroke(
                 mathfield.keystrokeBufferStates.length - (candidate.length - i);
             mathfield.keystrokeBuffer += c;
             mathfield.keystrokeBufferStates.push(mathfield.getUndoRecord());
-            if (
-                getInlineShortcutsStartingWith(candidate, mathfield.options)
-                    .length <= 1
-            ) {
-                resetKeystrokeBuffer = true;
-            } else {
+
+            if (shortcut) {
                 mathfield.resetKeystrokeBuffer({ defer: true });
             }
         }
--- a/src/editor-mathfield/mathfield-private.ts
+++ b/src/editor-mathfield/mathfield-private.ts
@@ -739,7 +739,7 @@ export class MathfieldPrivate implements Mathfield {
     resetKeystrokeBuffer(options?: { defer: boolean }): void {
         options = options ?? { defer: false };
         if (options.defer) {
-            if (this.options.inlineShortcutTimeout) {
+            if (this.options.inlineShortcutTimeout >= 0) {
                 // Set a timer to reset the shortcut buffer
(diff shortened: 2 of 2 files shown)


arnog:
OK, that's a tricky one.

The timeout check is actually correct, although it could be ">0" to be clearer. When the timeout is "0", that basically means "don't timeout", that is, regardless of the amount of time between keystrokes, consider them as candidate for an inline shortcut.

The other test (`if (getInlineShortcutsStartingWith`...) is also correct. It checks if there's a single matching shortcut: if so, the keystroke buffer can be reset right away: an inline shortcut has been found, and no future keystroke entry will cause us to revisit that.
If there's *more than one* shortcut (for example if there's an inline shortcut for 'sin' and another one for 'sinh'), then don't reset the keystroke buffer yet, only do a *potential* reset (= `defer: true`). If a subsequent keystroke matches the longer shortcut (i.e. "sinh") we'll undo the previous shortcut ("sin") and replace it with the new one.

The problem reported in #726 issue is caused by the `if (keystroke === '[Backspace]')` case. When backspace is pressed, the keystroke buffer states needs to be updated (which is done correctly). However, the keystroke buffer also needs to be adjusted: the code currently only drops the last keystroke in the keystroke buffer, but that's incorrect if the last keystroke caused a (temporary) match (for example, the "\times" was a match on "xx", so *two* characters should be removed, not just one). Not sure the best way to fix it, but it might be keep to keep track of the `keystrokeBuffer` in the `keystrokeBufferState` and to use that to restore the `keystrokeBuffer` when backspace is pressed.

alexprey:
Thanks for respond, we also found issue with `sin` and `sinh` and I'm continue fixing the issue.
When timeout set to 0 - currently buffer will not reset anymore when `defer` option are passed.

arnog:
Yes, that's the documented behavior. See [`InlineShortcutOptions`](http://cortexjs.io/docs/mathlive/#(%22options%22%3Amodule).(InlineShortcutsOptions%3Atype)):

>A value of 0 is the same as infinity: any consecutive character will be candidate for an inline shortcut, regardless of the interval between this character and the previous one.

Maybe that's not a very intuitive behavior. I suppose it would be reasonable to expect that "0" would be "immediate reset of the buffer", which would essentially turn off inline shortcuts.

alexprey:
Hm, thanks to highligh it, that was is not clear for me

PR #760, opened 2020-12-10
Fix issue with `undefined` object access

alexprey:
Fix unexpected exception with `undefined` object access when user starts typing with hidden virtual keyboard


arnog (approved):
looks good to me.

PR #761, opened 2020-12-10
Fix example pages errors

alexprey:


examples/basic/index.esm.html
```diff
@@ -32,25 +32,29 @@ <h1>MathLive Basic Example</h1>
     </footer>
 
     <script type="module">
-        import { makeMathField } from '/dist/mathlive.mjs';
+        import MathLive, { makeMathField } from '/dist/mathlive.mjs';
```
arnog, line 35: I don't think you need to include `MathLive` here, do you?

alexprey: That requires for `MathLive.latexToAST` method. Maybe we need to rewrite it to the new API from mathJSON, but I can not found how it can be done

arnog: ah, yes, that makes sense. That API will actually be imported from a separate MathJSON module. Coming soon, so this is good for now.

PR #771, opened 2020-12-22
Fix support for sup & sub for mathml output

alexprey:
Currently when forllowing latex text are used `\int_0^x` it produce the wrong output. This output dont have information about sup and sub for integral.


arnog:
Awesome. Thanks!

alexprey:
The same issue still keep for MathJson output. Im continue work on that. I create a new PR when done, I know that you plan to rework whole MathJson and for us mathml is more prioritized, so I done mathml first)

Issue #785, opened 2021-01-02
htmlData command


zoranpopovic:
Katex has a really handy latex command `\htmlData{foo=a, bar=b}{x}` that allows for storing data within expression components, which is crucial for keeping track of expression changes, and backend interpretation of the changed math content.  Would it be possible to add it to Mathlive?  I would imagine \\cssId and \\class could be implemented using htmlData...

arnog:
Yes, that's certainly possible.

Note that `\cssId` and `\class` are already supported.

arnog:
Fixed by #786.

PR #786, opened 2021-01-04
add Data command to add data properties to dom elements

zoranpopovic:
It basically mimics what was done for cssId, to add a command.   Addresses #785 

fields added have a form udata-* to avoid conflict with data- fields that mathlive already adds.

### DIFF
--- a/src/addons/definitions-metadata.ts
+++ b/src/addons/definitions-metadata.ts
@@ -188,6 +188,7 @@ metadata(
     '\\ttfamily',
     '\\class',
     '\\cssId',
+    '\\data',
   ],
   RARE,
   '{$0 Don Knuth}'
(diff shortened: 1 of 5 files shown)


arnog (changes requested):
Thank you for this contribution.

Looks good overall but I'd suggest these changes:
- rename the `\data` command to `\htmlData` for consistency with Katex
- rename the 'data' property to 'htmlData'
- output the data as 'data-' attributes rather than 'udata-'. The 'data-' attributes are valid HTML attributes, while 'udata-' is not valid HTML. The fact that Mathlive already use one 'data-' attribute should not be an issue.

zoranpopovic:
changes made.

arnog (approved):
Looks good. 
Thanks!

PR #796, opened 2021-01-15
Add display = [] to example code to make code snippet valid

rkeulemans:
Currently the code snippet is invalid, since the mandetory `display` key is not included. I would argue that is is a good idea to have only working code in the documentation.


arnog (approved):
Good catch. Thanks!

PR #858, opened 2021-03-15
fix for cssData

zoranpopovic:
single space was missing, making data-* not work at all.


zoranpopovic:
@arnog any chance this can make it soonish in a new release.  Also i noticed that the new replace() method has a leftover console.log  testing entry.

arnog (approved):

PR #905, opened 2021-04-16
Allow appending the virtual keyboard to an arbitrary element

djankovic:
I added a new option, `virtualKeyboardAppendTarget`, that enables customization of where in the DOM the virtual keyboard gets inserted. 

The use case for this is having a working keyboard when [full screening DOM elements](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API) that contain MathLive, eg.

```html
// ...

<div id='container'>
  <div id='math-field'></div>
</div>
<button id='toggle-fullscreen'>Toggle fullscreen</button>

<script type='module'>
  import MathLive from 'mathlive'

  const container = document.getElementById('container')
  document.getElementById('toggle-fullscreen')
    .addEventListener('click', () => container.requestFullscreen())

  const mathField = document.getElementById('math-field')  
  MathLive.makeMathField(mathField, {
    virtualKeyboardMode: 'onfocus',
    virtualKeyboardAppendTarget: container
  })
</script>

// ...
```

Currently, on `master`, the keyboard isn't visible after clicking the Toggle fullscreen button and focusing the MathLive input because it renders outside of `#container` which is fullscreened. This commit fixes that.

### DIFF
--- a/src/editor/options.ts
+++ b/src/editor/options.ts
@@ -204,6 +204,10 @@ export function update(
         }
 
         break;
+      case 'virtualKeyboardAppendTarget':
+        result.virtualKeyboardAppendTarget =
+          updates.virtualKeyboardAppendTarget;
+        break;
       case 'onBlur':
       case 'onFocus':
       case 'onContentWillChange':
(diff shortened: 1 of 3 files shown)


arnog:
Sounds good. Thanks for the contribution!

PR #953, opened 2021-05-27
Fix loosing brackets in mathfield for htmlData class cssId tex commands

eds007:
### Description
When typing `\htmlData{test:test}{x=0}` in mathfield than adding something else - tex code looses brackets in second part of command. For example ` \htmlData{test:test}{x=0}` => `\htmlData{test:test}x=0somethingelse`. `\class`, `\cssId` has the same behaviour.
### Expected result
`\htmlData{test:test}{x=0}` => `\htmlData{test:test}{x=0somethingelse}`

### DIFF
--- a/src/core-atoms/group.ts
+++ b/src/core-atoms/group.ts
@@ -85,13 +85,13 @@ export class GroupAtom extends Atom {
     }
 
     if (this.htmlData) {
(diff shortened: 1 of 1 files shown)


arnog:
Great catch. Thanks for the PR!

PR #986, opened 2021-06-05
Fixed expression order for support on regex unicode property escape

ChrBrkmn:
Stumbled upon this when getting a syntax error (`invalid identity escape in regular expression`) in Firefox ESR 68.12.0. 

`supportRegexPropertyEscape` returns, that this browser version does not support property espaces, but uses it anyway.
Switched the expressions, so it is only being used when the browser actually supports it.


arnog:
Oops... Thanks for the PR!

PR #1005, opened 2021-06-10
test: make tests run again

AceGentile:
fix broken imports and mock stylesheet imports
in order to make tests be able to run again


arnog:
Thanks for this PR! The tests were assuming that a build had been performed before running, but with your changes the tests can run without a build.

PR #1112, opened 2021-08-04
fix: navigator access on the server

illright:
Seems like the optional chaining operator wasn't enough to guard against `ReferenceError` on the server. This PR will add guards wherever `navigator` may be unsafely accessed.

Will also do the same for `window`.

### DIFF
--- a/src/common/capabilities.ts
+++ b/src/common/capabilities.ts
@@ -31,6 +31,10 @@ export function osPlatform():
   | 'ios'
   | 'chromeos'
   | 'other' {
+  if (!isBrowser()) {
+    return 'other';
+  }
+
   const platform = navigator.platform;
 
   if (/^(mac)/i.test(platform)) {
--- a/src/editor/keybindings.ts
+++ b/src/editor/keybindings.ts
@@ -11,13 +11,13 @@ import {
 } from './keyboard-layout';
 import { REVERSE_KEYBINDINGS } from './keybindings-definitions';
 import type { ParseMode } from '../public/core';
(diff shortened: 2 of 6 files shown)


arnog:
Thanks for this PR.
Can you expand a bit on what this fixes? Why wasn't the optional chaining operator not working?
I can understand the fix in `capabilities.ts`, and I can see the check line 338 in `options.ts` is insufficient (is should have a nullish operator i.e. `?? ''`, but I don't understand the others.

illright:
Hi, sure. There's a difference in accessing undefined fields and undefined variables. For instance:
```js
const empty = {};
if (empty.field != null) {
  console.log(empty.field);  // perfectly safe, ?. could also work
}
```

```js
// let's pretend that the variable `undef` is conditionally defined
if (undef != null) {
  console.log(undef);  // will throw a ReferenceError
}

if (typeof undef !== 'undefined') {
  console.log(undef);  // perfectly safe
}
```

Since `window` and `navigator` are undefined variables, not simply fields, they require special treatment. See [this Stack Overflow answer](https://stackoverflow.com/a/519157/5726823).

illright:
I am not 100% sure I've covered all the necessary places of access with guards, those that would be reachable by mere imports, if you could check this, I would highly appreciate :)

arnog:
Ok, got it. Thanks, I’ll merge it and look for other potential call sites.

illright:
When do you plan to roll out the next release?

arnog:
I just publish mathlive@0.69.5

PR #1147, opened 2021-08-24
Reduce mathfield min height

stefnotch:
This mostly prevents something I've asked about on Gitter

> I realize that this is a bit of an application specific question, but is there a way of making the `ML__fieldcontainer__field` take up the size of its parent (`ML__fieldcontainer`)?
> Currently, it's possible to have a field that's a bit smaller than its parent, leading to the user pressing in the mathfield and wondering why the caret doesn't get placed anywhere
> image
>

### DIFF
--- a/css/mathfield.less
+++ b/css/mathfield.less
@@ -60,7 +60,7 @@
   flex-flow: row;
   justify-content: space-between;
   align-items: flex-end;
-  min-height: 39px; /* Need some room for the virtual keyboard toggle */
+  min-height: 12px; /* Need some room for the virtual keyboard toggle */
 
   /* Prevent the browser from trying to interpret touch gestures in the field */
   /* "Disabling double-tap to zoom removes the need for browsers to


css/mathfield.less
```diff
@@ -60,7 +60,7 @@
   flex-flow: row;
   justify-content: space-between;
   align-items: flex-end;
-  min-height: 39px; /* Need some room for the virtual keyboard toggle */
+  min-height: 12px; /* Need some room for the virtual keyboard toggle */
```
arnog, line 63: Doesn't this cause a problem, as per the comment on that line? (that is the height must be large enough to have room for the virtual keyboard toggle)

stefnotch: That's true, I haven't properly tested it with the virtual keyboard toggle. My app doesn't have one after all.

I changed the approach to always make the `ML__fieldcontainer__field` as large as the `ML__fieldcontainer`. Since the `ML__fieldcontainer` contains both the `ML__fieldcontainer__field` and the virtual keyboard toggle, this works out fine.

PR #1156, opened 2021-08-30
fix: typo in virtual keyboards MF property

yakovlev-alexey:
Stumbled upon this typo while debugging virtual keyboards in my project. Pretty sure there should be no boars involved 😅 🐗


arnog:
😀 Great catch, thanks!

PR #1185, opened 2021-09-24
fix #951 mathML sup sub error

### LINKED ISSUE #951: Exponent expressions with a number as the base lose the exponent when convertd to MathML.
## Description

I'm using the 0.63.1 tagged release. When trying to enter an expression that consists of a exponent with a constant base, the exponent is dropped and the resulting expression is just the base. So, for example, if the expression is 2^3, when converted to MathML the result is 2. 

## Steps to Reproduce

Check the results of MathLive.convertLatexToMathMl for an exponential expression with a constant base, such as 2^3 or any similar expression. 

### Actual Behavior

`<mn>2</mn>`

image

### Expected Behavior

`<msup><mi>2</mi><mn>3</mn></msup>`

### Environment

**0.63.1**

**Windows 10**

**Firefox 88.0.1**

eds007:


arnog:
Looks good. Good catch, thank you!

PR #1197, opened 2021-09-29
virtual-keyboard-utils: fix broken import

AceGentile:
fix broken import on virtual-keyboard-utils
throwing an error on install


arnog:
Sorry about that. I had an uncommitted fix for this, but you beat me to it! Thanks.

PR #1199, opened 2021-09-29
start: use serve-http instead of http-serve to fix node 14 bug w/ linux

AceGentile:
serve-http has an issue with a node >= 14 fs.watch breaking
change on linux causing the server to crash.
on efea7a7a0080870712cdc81a394435b4953e38d1, while introducing
estrella, serve-http was dismissed in favor of http-serve.

In order to be able to run a server on linux machines,
let's re-introduce http-server

rif. https://github.com/rsms/serve-http/issues/6


arnog:
LGTM. Thanks for the PR!

PR #1200, opened 2021-09-29
Fix #1087 - math-field scrolling on input inconsistent behavior

### LINKED ISSUE #1087: Virtual keyboard: caret not scrolling when inserting symbols
## Description
Using the virtual keyboard as input, I found that, in case of text overflow, the caret is not always visible depending on which keys are pressed.
In the screen recording below I've inserted some _e_s until I overflow the input, then the caret is visible only after I insert a 9.

https://user-images.githubusercontent.com/902985/125607949-5980975c-9157-46af-8ee4-76854d5e6335.mp4

Demo here: https://purple-aged-vanilla.glitch.me

## Steps to Reproduce

1. focus the input
2. insert some symbols like e, i or π using the virtual keyboard until the text overflow
3. **the input is not scrolled left to keep the caret visible**
4. press a number on the virtual keyboard
5. **the field content is scrolled left and the caret is visible again** 

### Expected Behavior
The caret should always be visible after the input overflow.

### Notes

I've looked in to it a bit and it seems that there is a difference in the scroll handling between the insertion via keyboard and the `typedText` API and the insertion of LaTeX using the `insert` API. 

### Environment

**MathLive version**  0.69.4

**Operating System** macOS

**Browser** Safari, Chrome, Firefox

baldoalessandro:
Hello 👋 I was able to tackle part of the problem discussed in #1087.
This change enables the field to scroll after input for command that are using the `performWithFeedback` method, with the same behavior as keybindings and typedText.

This alone will allow to move the caret in the math-field using the left/right buttons in the virtual keyboard with the same experience as using the arrow keys on the physical keyboard.

I'm still figuring out how to extend this fix to buttons that use the `insert` method.
Any help in this direction would be wonderful!

### DIFF
--- a/src/editor-mathfield/mathfield-private.ts
+++ b/src/editor-mathfield/mathfield-private.ts
@@ -972,6 +972,10 @@ export class MathfieldPrivate implements Mathfield {
         void this.keypressSound?.play().catch(console.warn);
       }
 
+      if (options.scrollIntoView) {
+        this.scrollIntoView();
+      }
+
       if (s === '\\\\') {
         // This string is interpreted as an "insert row after" command
         addRowAfter(this.model);
--- a/src/editor/commands.ts
+++ b/src/editor/commands.ts
@@ -218,7 +218,9 @@ export function performWithFeedback(
     mathfield.keypressSound?.play().catch(console.warn);
   }
 
-  return mathfield.executeCommand(selector);
+  const result = mathfield.executeCommand(selector);
+  mathfield.scrollIntoView();
+  return result;
 }
 
 register({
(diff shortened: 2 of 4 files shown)


arnog:
Looks good.
For the insert case, here's what I think needs to be done:
- In `InsertOptions` in `mathfield.ts` add a `scrollIntoView` boolean property
- in `insert()` in `mathfield-private.ts`, after (or before) `if (options.feedback)` add a `if (options.scrollIntoView)` that calls `scrollIntoView()`.
- in `virtual-keyboard-utils.ts`, search for `'insert'` (there's two call sites). Add `scrollIntoView: true` in the selector arguments (next to `feedback: true`.

I believe that should take care of  it.

Issue #1195, opened 2021-09-27
Wrong MathMl output for sub/superscript equations, using group brackets


eds007:
## Description

Wrong mathMl output for equations with groups in subsupersript.

## Steps to Reproduce

[Provide steps that are specific and repeatable]

1. type x^{{x}^1}_2
2. get mathML putput
3. Try to render it on https://www.mathjax.org/#demo or https://www.mathmlcentral.com/Tools/FromMathML.jsp
4. type x^{x^1}_2
5. get mathML output
6. do 3-rd step
 
### Actual Behavior

In case 'x^{{x}^1}_2' mathlive generates wrong mathML code. No one online decoder can render it.
 Ethalon convertion latex - mathml - latex doesn`t work.

### Expected Behavior

 x^{x^1}_2  and x^{{x}^1}_2 gives  equal mathML code.
 Convertion latex - mathml - latex works correct
### Environment

**MathLive version**
0.68.1
0.69.7

**Operating System**
windows 10

**Browser** 
Chrome
![Screenshot_4](https://user-images.githubusercontent.com/83068117/134878266-1f67cae9-a152-48d2-845d-82c83e41da94.png)
******************************************
![Screenshot_1](https://user-images.githubusercontent.com/83068117/134878271-08bdaeef-b5b7-4e74-b12e-8d331f2793ec.png)

arnog:
Fixed by PR

arnog:
Re-opening. The PR did not work.

arnog:
Note: there's a `<mrow>` missing. Should output: `<msubsup><mi>x</mi><mn>2</mn><mrow><mi>x</mi><msup><mn>1</mn></msup></mrow></msubsup>`

PR #1270, opened 2021-12-13
fix(1195): correct parsing of subgroup elements

i3ima:


i3ima:
Fixes #1195

arnog:

arnog:
Upon closer examination, the PR has a problem: it ends up creating an infinite loop when running the demo (`npm run start`). Reverting it for now.

PR #1299, opened 2022-01-09
added commnad "\HTMLStyle" to support html styling

neokazemi:
HTML styling is supported in Katex and I think this will be a good idea to be supported here too. [Refrence](https://katex.org/docs/supported.html#html).

I made the necessary changes and tested them.

Thanks.


arnog:
Looks good to me.  Thank you for your contribution!

Issue #1300, opened 2022-01-09
roundedbox not displaying


Zahara-Nour:
## Description

`\enclose{roundedbox}`doesn't display a box.

## Steps to Reproduce

Try with testing playground, `\enclose{roundedbox}[2px solid rgba(0, 255, 0, .8)]{\frac{1}{2}}`
<img width="847" alt="Capture d’écran 2022-01-09 à 22 07 05" src="https://user-images.githubusercontent.com/32236491/148696950-ce5a796d-01cb-4b5d-825b-09cd059d26f4.png">

### Environment

**MathLive version**  commit 7204ac26b39d6bf65ad926ef497be209fd10f7cb

**Operating System** macOS BigSur 11.6

**Browser** Chrome, Firefox

Zahara-Nour:
I just noticed that the latex output is not correct : the border width is `2 pxpx `although I checked I entered `\enclose{roundedbox}[2px solid rgba(0, 255, 0, .8)]{\frac{1}{2}}`.

Zahara-Nour:
Found it ! Will make a PR.

PR #1309, opened 2022-01-15
Fix : roundedbox not displaying #1300

Zahara-Nour:
The border box didn't display.


arnog:
Good catch! Thanks for the PR!

Issue #1319, opened 2022-01-27
Different behavior when toggling `readonly` or `read-only` programmatically 


LuisMesa:
## Description

Toggling `readonly` or `read-only` attributes in a `math-field` element is giving different results. Changes to the `readonly` attribute are observed(`observedAttributes`) but changes to `read-only` are not. For example, a `math-field` where `read-only` was just added would allow users to edit its content.

In addition, a `math-field` where `readonly` or `read-only` were just added would show a virtual keyboard on focus (if it has `virtualKeyboardMode: "onfocus"`). This allows users to modify the content if they press certain keys in the keyboard, for example `sqrt`.

Finally, I have noticed that when the `readonly` attribute is added, the `read-only` attribute is also added somehow, but when `readonly` is removed, `read-only` stays.

You can see here the behavior of a `math-field` after toggling `read-only` or `readonly`:
https://codepen.io/LuisMesa/pen/GROgOZz?editors=1010

Also, you can find here some videos showing what is described above.

https://user-images.githubusercontent.com/14155607/151276682-6043e35e-c2ef-4f88-a975-c75746211bec.mov

https://user-images.githubusercontent.com/14155607/151278272-5e6d26e9-4386-4ad3-a491-c2bba54e3647.mov

## Steps to Reproduce

1. Go to https://codepen.io/LuisMesa/pen/GROgOZz?editors=1010
2. Click the 'Toggle read-only' button.
3. Click the `math-field`
4. Start typing.

### Actual Behavior

The `math-field` is editable even though it has the `read-only` attribute. Also, the virtual keyboard is opened.

### Expected Behavior

The `math-field` should not allow users to edit it. Also, the virtual keyboard should not be opened.

### Environment

**MathLive version** 0.69.9

**Operating System** macOS Big Sur 11.3.1

**Browser** Chrome 97.0.4692.71, Safari 14.1

arnog:
Fixed with PR above

PR #1320, opened 2022-01-27
fix: properly observe changes to the `read-only` attribute

LuisMesa:
Before, changes to the `readonly` attribute were observed but changes to `read-only` were not. This was causing wrong behaviors when adding `read-only` to a `math-field` element, for example it was allowing users to modify the content of a `math-field` after adding the `read-only` attribute programmatically.
You can see here the behavior of a `math-field` after toggling `read-only` or `readonly`:
https://codepen.io/LuisMesa/pen/GROgOZz?editors=1010

https://user-images.githubusercontent.com/14155607/151276682-6043e35e-c2ef-4f88-a975-c75746211bec.mov

This is a related issue I just created: https://github.com/arnog/mathlive/issues/1319
This PR will prevent users from typing in the `math-field` after `read-only` was added.

### DIFF
--- a/src/public/mathfield-element.ts
+++ b/src/public/mathfield-element.ts
@@ -603,6 +603,7 @@ export class MathfieldElement extends HTMLElement implements Mathfield {
       ...Object.keys(MathfieldElement.optionsAttributes),
       'disabled', // Global attribute
       'readonly', // A semi-global attribute (not all standard elements support it, but some do)
+      'read-only',
     ];
   }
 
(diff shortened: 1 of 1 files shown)


arnog:
Thanks for the PR and for documenting the issue.

It seems reasonable,  so I'm going to merge it in.

Unfortunately, `readonly` is a DOM-standard attribute, while `read-only` follows the conventions of the other custom attributes, so both are kinda needed. Getting them to behave consistently, and logically, has been a challenge...

Issue #1024, opened 2021-06-19
\ne ,\neq and \not= are not working.


faxinwang:
## Description

\ne ,\neq and \not= are not working. 

input: A\ne B
output: AB
expected:A\ne B

input: A\neq B
output: AB
expected: A\neq B

input A\not=B
output: A=B
expected:A\not= B

Note that the editor dispalying the right content, only the output latex value is not as expected.

### Environment

**MathLive version** 0.69.3

**Operating System** [Windows 10]

**Browser** [ Chrome]

stefnotch:
Additional Info:
If one types `A \ne B`, the copied MathJson ends up being: `["Multiply","A","B"]`. This seems to be related to the issue

arnog:
Ah, yes, that's probably the problem...

faxinwang:
Do you mean I just need to upgrade mathlive to 0.69.4 ?

I have upgraded to 0.69.4, but the problem still exists.

faxinwang:
Hi, is there any progress on this problem? I have tried typing A \ne B on this page(https://cortexjs.io/mathlive/demo/), the output is still AB。

arnog:
I believe this was a duplicate of #1222, which is now fixed.

PR #1336, opened 2022-02-10
core-definitions: fix \ne \neq rendering issue

AceGentile:
fix `\ne` `\neq` disappearing operator bug documented in issue #1024

`\ne` and `\neq` operators' `createAtom` function is returning a `GroupAtom` class instance with an empty command value.
This causes `Parser`'s `parseCommand` method to populate `verbatimLatex` value with an empty string.
The empty `verbatimLatex` value causes a malfunction in `Atom`'s static `serialize` method removing `\neq` and `\ne` operators from displayed latex.

### DIFF
--- a/src/core-atoms/group.ts
+++ b/src/core-atoms/group.ts
@@ -30,6 +30,7 @@ export class GroupAtom extends Atom {
       style?: Style;
       captureSelection?: boolean;
       serialize?: (atom: GroupAtom, options: ToLatexOptions) => string;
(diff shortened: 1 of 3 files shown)


AceGentile:
Not 100% sure this is the correct bugfix for this issue since I'm not familiar with the codebase, but seems to work on my code :)

arnog:
Yep, I think that makes sense. Thanks for the PR!

PR #1365, opened 2022-03-05
fix: link to Contributor Guide

manstie:
Direct the `Contributor Guide` link to the correct file path.


arnog:
Thanks for the fix!

PR #1384, opened 2022-03-13
fix: add appropriate gap to aligned environment

manstie:
Before:
image

After:
image

LaTeX: `\begin{aligned}x^2 & =12 \\ x & =\sqrt{12}\end{aligned}`


arnog:
Looks good to me, and indeed more closely matches LaTeX output. Thanks for your contribution!

PR #1386, opened 2022-03-13
fix: Fixed #1362 and #726

### LINKED ISSUE #726: 'x' symbol pressed several times leads to strange behavior

### LINKED ISSUE #1362: Compound operators linger when backspaced
## Description

When a "compound operator" - an operator made up of two or more keystrokes (e.g. `>` and `=` to make `\ge`/`≥`) is backspaced, the backspace removes the whole operator (visually) but effectively only the last operand is removed.

## Steps to Reproduce

1. Type in `x >`
2. Type a `=` to turn the symbol into a `\ge`
3. Backspace the combined operator `\ge`
4. Type a `=` to attempt to create an equal operator
5. Notice a `\ge` operator was created (as the `>` operator lingers)

Same thing happens with typing:
`>= {backspace} >` = `\gg` instead of `>`
`=> {backspace} =>` = `\Longrightarrow` instead of `\Rightarrow`
`==> {backspace} >` = `\Longrightarrow` instead of `>`
etc...

### Environment

**MathLive version**
commit 2dc8bf624191d5103c150779cf84d9b36d3571cc (tag: v0.69.10, origin/master, origin/HEAD, master)

**Operating System**
Windows 10 Pro
10.0.19042 Build 19042

**Browser**
Issue is not browser dependent - tested in Firefox, Chrome, Edge Chromium

manstie:
I think this is an appropriate fix for these issues. I don't think it breaks anything else.

- `sin` -> `sinh` still works
- `xx` followed by 1 more `x` displays `×x` instead of `x×`
- all combinations of `>` and `=` and backspacing act as expected

Let me know if I missed anything.

closes #726
closes #759
closes #1362

### DIFF
--- a/src/editor/shortcuts.ts
+++ b/src/editor/shortcuts.ts
@@ -15,12 +15,9 @@ export function getInlineShortcutsStartingWith(
 ): string[] {
   const result: string[] = [];
 
-  for (let i = 0; i <= s.length - 1; i++) {
-    const s2 = s.slice(Math.max(0, i));
-    for (const key of Object.keys(config.inlineShortcuts)) {
-      if (key.startsWith(s2) && !result.includes(key)) {
-        result.push(key!);
-      }
+  for (const key of Object.keys(config.inlineShortcuts)) {
+    if (key.startsWith(s)) {
+      result.push(key!);
     }
   }
 


arnog:
Hmmm... I think this is OK. At least I can't think of a case where this would be worse than before. Thanks for looking into this!

PR #1389, opened 2022-03-19
feat: allow navigation of ArrayAtom with up and down keys

manstie:
Not sure if this is the way you would have liked this implemented, but nevertheless here it is.

A matrix to test on:
`\begin{bmatrix}a & bbbbb \\ c & dddd \\ e & fff\end{bmatrix}`

I have a few questions regarding the code which I hope you could assist.

```
      // doesn't highlight
      let newSelection: Selection;
      if (targetSelection < left) {
        // extending selection upwards
        newSelection = {
          ranges: [[targetSelection, right]],
          direction: 'backward'
        };
      } else {
        // reducing selection upwards
        newSelection = {
          ranges: [[left, targetSelection]],
          direction: 'forward'
        };
      }
      model.setSelection(newSelection);
```
This code ends up not highlighting the selection, and if you look at the selection details it's only updating the "Position" value - it gets corrected when you hold shift to continue your selection and press left or right.
image
Is this a bug in the `setSelection` function or am I using it incorrectly?

There are also other segments of code I tried:
```
      // does highlight, has direction error
      // model.extendSelectionTo(
      //   targetSelection < left ? right : left,
      //   targetSelection
      // );
```
This displays the selection properly but has a quirk with the direction, e.g. if you start selecting from left to right the "fff" on the bottom right row, and then use `shift + up arrow` to start highlighting above, any subsequent `shift + left/right` will deselect the parts of the selection that were left to right (forward). Unsure if this is also a bug or using the function incorrectly.

And then there's this interesting quirk:
```
      // doesn't highlight, doesn't continue selection
      // model.extendSelectionTo(
      //   targetSelection,
      //   targetSelection < left ? right : left
      // );
```
If you swap the parameters from the other example, it just doesn't work at all. I can't figure out why. The first thing it does in the function is `normalizeRange` which should make the order of the parameters indifferent to the function, so I can't figure out why it behaves differently. Another bug?

There is one feature which I would like to add and that is remembering the index from which you came from. e.g. With the cursor on the right of the top-right row which contains `bbbbb`, if I press down, down, up, up (with no left or right in-between) I believe the desired cursor position would still be at the right of all the b's. If I were to add a variable to track something like that, would it best be placed in the model or in the atom (or even local to the file)?

Thanks, I look forward to your feedback.

### DIFF
--- a/src/editor-model/commands.ts
+++ b/src/editor-model/commands.ts
@@ -1,8 +1,9 @@
 import type { ModelPrivate } from './model-private';
+import { ArrayAtom } from '../core-atoms/array';
 import { LatexAtom } from '../core-atoms/latex';
 import { TextAtom } from '../core-atoms/text';
 import { LETTER_AND_DIGITS } from '../core-definitions/definitions';
-import { Offset } from '../public/mathfield';
+import type { Offset, Selection } from '../public/mathfield';
 import { getCommandSuggestionRange } from '../editor-mathfield/mode-editor-latex';
 
 /*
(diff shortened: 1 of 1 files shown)


arnog:
Thank you so much for this PR! I apologize, I had missed it and just saw it today.

Looks good overall. Certainly an improvement over the current situation.

> Is this a bug in the setSelection function or am I using it incorrectly?

You are using it correctly, and `setSelection` is the right method to use. It's just that this particular case is not correctly handled by `setSelection`. There is however a `@todo` tag (see `model-private.ts/setSelection:140`. The basic idea is to set the selection to an array of ranges, one for each column. There are some things to be worked out, like what happens when the column is "empty". Another, simpler, option, would be to set the range to simply encompass all the index, even if they include some that are in other cells.

> Unsure if this is also a bug or using the function incorrectly.

I'm  pretty sure it's the correct behavior for `extendSelectionTo`. As you note it has "quirks" when used programmatically, but it's really intended to be used as a direct response to a user input where those quirks are more in line with user expectations.

> If I were to add a variable to track something like that, would it best be placed in the model or in the atom (or even local to the file)?

Definitely not local to the file. I think the right place for it would be in the atom for the cell.

Again, thank you and my apologies for the delay.

PR #1416, opened 2022-04-06
Group atoms inside of parentheses as implicit args (#1415)

### LINKED ISSUE #1415: Parenthesized expressions aren't implicit arguments with `smartFence: false`
## Description

The "implicit argument" feature for division can fail for Parenthesized expressions, leaving a blank fraction:

CleanShot 2022-04-05 at 15 49 12

This happens either if:
1. `smartFence` is set to `false`
2. You delete the closing parenthesis and re-add it. (see following video)

https://user-images.githubusercontent.com/446007/161867858-43221875-8ff3-40f4-916b-d2afa9f748ed.mp4

## Steps to Reproduce

1. Set `smartFence: false` on a MathField.
2. Type in a parenthesized expression like `(1+x)`.
3. Press the `/` key.

OR:

1. Type in a parenthesized expression like `(1+x)`.
2. Delete the `)`
3. Re-enter the `)`
4. Press the `/` key

### Actual Behavior

The field should contain `(1+x)/[]`

### Expected Behavior

The field contains `(1+x) []/[]`

### Environment

**MathLive version**: 0.70.0 (but happens at least as far back as 0.69.8)

**Operating System** macOS

**Browser** Safari

### Notes

It seems that this is happening because MathLive represents `\left(` differently from `(` internally. `\left(` and `\right)` pairs are represented as `leftright` atoms with the contents as children, and `(``)` pairs are represented as individual `mopen` and `mclose` atoms.

@arnog -- I can take a stab at this, but could use your opinion; it seems like the two approaches to solving this are:
1. Represent normal paired parentheses as `leftright` atoms
2. [Make `getImplicitArgOffset` ](https://github.com/brilliantorg/mathlive/blob/0bd81c362981f00af2bb6d4212cdaf00d0633ea6/src/editor-mathfield/mode-editor-math.ts#L511) more sophisticated and have it try to pair up right/left delimiters as it scans right-to-left.

bengolds:
# What
This fixes #1415 and keeps the contents of parentheses together when you press the division key.

Behavior in a few cases can be seen here:
https://user-images.githubusercontent.com/446007/162076602-12b883c3-6093-4d42-95b0-9adae3c92b5d.mp4

 # How
When calculating what the "implicit argument" to division is, we follow this algorithm:

1. Consume atoms until you hit some sort of binary operator OR until you hit a closing delimiter.
2. If you hit a closing delimiter, consume atoms until that delimiter has been closed.

We keep track of which delimiters we've seen so far so that nested delimiters work correctly.

# Notes
If the closing delimiter is not matched up, we consider the entire preceding expression part of the implicit argument. There may be better behavior here -- thoughts, @arnog?

We look for exact matches of delimiters, not visual matches. So, an expression like `1 + \left(x)` might _look_ like it should consider `\left(x)` as the implicit argument, but it will actually treat the whole thing as the implicit argument since the right parenthesis is never matched. @arnog -- do you have a dictionary of "visually similar" delimiters yet, akin to `RIGHT_DELIM` but more general?

### DIFF
--- a/src/core/delimiters.ts
+++ b/src/core/delimiters.ts
@@ -54,6 +54,12 @@ export const RIGHT_DELIM = {
   '\\lmoustache': '\\rmoustache',
 };
 
+export const LEFT_DELIM =
+  Object.fromEntries(
+    Object.entries(RIGHT_DELIM)
+      .map(([leftDelim, rightDelim]) => [rightDelim, leftDelim])
+  );
+
 function getSymbolValue(symbol: string): number {
   return (
     {
--- a/src/editor-mathfield/mode-editor-math.ts
+++ b/src/editor-mathfield/mode-editor-math.ts
@@ -17,7 +17,7 @@ import {
 
 import type { Style } from '../public/core';
 import { LeftRightAtom } from '../core-atoms/leftright';
-import { RIGHT_DELIM } from '../core/delimiters';
+import { RIGHT_DELIM, LEFT_DELIM } from '../core/delimiters';
 import {
   contentDidChange,
   selectionDidChange,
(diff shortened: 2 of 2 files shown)


arnog:
This looks good to me.

I think the behavior is reasonable when there is no match.

And unfortunately, there's nothing better than `RIGHT_DELIM`. It can get quite complicated, in theory, although in practice I think those cases are rare, i.e. `\left\lparen...\right)`.

Thanks for the PR!

PR #1431, opened 2022-04-28
fix: typo in auto-render.ts

EaswarRaju:
fix typo in code

static math expressions not rendering inline (textstyle) when latex code is wrapped inside `<script type='math/tex; mode=text'>` . The issue is the result of a minor typo in code which this Pull Request aims to fix


arnog:
That's a good catch. Thank you!

PR #1457, opened 2022-05-18
Fix shared keyboards in the same frame

dannygreg:
Whilst pursuing more control over keyboard animations I noticed that shared keyboards were infinite looping if the `math-field`s were in the same frame as the keyboard itself. 

This PR breaks that infinite loop with a rather ugly check to see if we're in an iFrame or not before re-sending message events that are being handled in two places. Definitely open to ideas on a better fix - I was thinking that different events may make sense but it's somewhat tangled in the command "target" logic.

There are a couple of additional pieces here that I thought may still be useful, although ancillary to the fix itself:

* A type definition for the virtual keyboard modes
* More forgiving comparison of the mode strings 
    - As these can be declared in HTML, it's not guaranteed they are type checked and… *cough* someone *cough* may use an unexpected case in their HTML which confuses the comparison
* Avoid hiding and re-showing the keyboard if it's already visible and asked to show


arnog (changes requested):
Thanks for these changes. I've got three comments:
- you call `toLowerCase()` on the value of `virtualKeyboardMode`. That's not an unreasonable extra-safety measure, but I would suggest to do this when `virtualKeyboardMode` is set, rather than each time it is read. The setting is done in `optioons.ts`, in `update()`.
- I'm  not sure of the rationale for the changes in `virtual-keyboard-commands.ts`. Is this to suppress an undesirable animation? One side effect of it is that the `virtual-keyboard-toggle` event is no longer dispatched if the state of the keyboard doesn't effectively change. Maybe that's OK, although knowing that an attempt to change the visibility of the keyboard was made might be useful too.
- I don't understand the changes in `remote-virtual-keyboards.ts`. You recognize and intercept the command if the remote virtual keyboard is in the same window as the mathfield, but you don't handle the command. Isn't that a problem?

dannygreg:
> * you call `toLowerCase()` on the value of `virtualKeyboardMode`. That's not an unreasonable extra-safety measure, but I would suggest to do this when `virtualKeyboardMode` is set, rather than each time it is read. The setting is done in `optioons.ts`, in `update()`.

Great shout - updated to reflect.

> Maybe that's OK, although knowing that an attempt to change the visibility of the keyboard was made might be useful too.

Yeah agreed, considering I'm now suppressing the animation another way I don't mind just removing this. 

> I don't understand the changes in `remote-virtual-keyboards.ts`. You recognize and intercept the command if the remote virtual keyboard is in the same window as the mathfield, but you don't handle the command. Isn't that a problem?

This is the crux of the PR. Both `VirtualKeyboardDelegate` and `RemoteVirtualKeyboard` register for `message` events. When this was implemented there appears to have been an assumption made that a `VirtualKeyboardDelegate` would always be in a child frame of `RemoteVirtualKeyboard`. Therefore the previous behaviour was to re-send the message to the parent window.

However, if there is no parent window, [`window.parent` returns `window`](https://developer.mozilla.org/en-US/docs/Web/API/Window/parent)… which is a bit of an odd API decision but because of this it causes the message to be sent to the same window again - creating an infinite loop. 

The fix here is to avoid resending those messages if they would be reposted to the same window - which breaks the infinite loop. Hence the checks to see if `window === window.parent`. An alternate fix would be split the subscriptions so there are two different events but that has its own untidiness.

I'll fix up these conflicts and then send this back for review.

src/editor-mathfield/remote-virtual-keyboard.ts
```diff
@@ -50,7 +50,6 @@ export class VirtualKeyboardDelegate implements VirtualKeyboardInterface {
     this.targetOrigin = options.targetOrigin ?? window.origin;
     this.originValidator = options.originValidator ?? 'same-origin';
     this._mathfield = options.mathfield;
-    this.enable();
```
arnog, line 53: Isn't this call to `enable()` needed here?

arnog (commented):
Looks good. Just had a question about the removable of `this.enable()` in the `VirtualKeyboardDelegate` constructor.

src/editor-mathfield/remote-virtual-keyboard.ts
dannygreg: Good question! No, as it's enabled whenever a `math-field` gets focus. In fact, with this call I found that the keyboard will send commands to the incorrect fields.

PR #1460, opened 2022-05-19
Allow options changes to visible virtual keyboards

dannygreg:
Currently if options are set on a visible keyboard it is implicitly removed from the DOM and not re-shown.

This change instead rebuilds the keyboard to reflect the new options if it is visible.


arnog (approved):
Looks good to me.

PR #1463, opened 2022-05-23
Allow `setOptions` on remote keyboards

dannygreg:
This was the last missing piece for a "persistent" keyboard.

Instead of defining one set of options that must be shared between all fields - this makes the shared keyboard's behaviour match that of regular keyboards (from a caller's perspective): as a `math-field` gets focus the keyboard matches the options that have been set on the field itself.

Specifically this will allow different keyboard types to be shown for different fields whilst still giving the caller control over when to hide and show the keyboard.

The only limitations are not being able to change `createHTML` or the keyboard container itself, which feel very reasonable.

---

This also fixes a regression where the keyboard would not animate for shared keyboards due to `is-visible` being incorrectly set.


arnog:
Looks good to me!

PR #1520, opened 2022-06-26
prevent fonts loading multiple times

EaswarRaju:
`isBrowser() && !/firefox|safari/i.test(navigator.userAgent);`
The above test for browser in `supportLocalFontEnumeration` method is failing as Chrome and Microsoft Edge also has the word "Safari" in navigator.userAgent

The failed condition forcing the fonts to load multiple times in all major browsers, and resulting in static math to flicker each time mathfield is focused and virtual keypad is toggled to `showVirtualKeyboard`

This Pull Request aims to add alternate approach to detect loaded fonts in browser, and is tested to be working on latest versions of Chrome, Firefox, Safari and Microsoft Edge

### DIFF
--- a/src/core/fonts.ts
+++ b/src/core/fonts.ts
@@ -1,4 +1,4 @@
-import { isBrowser, supportLocalFontEnumeration } from '../common/capabilities';
+import { isBrowser } from '../common/capabilities';
 import { resolveRelativeUrl } from '../common/script-url';
 import { ErrorListener, MathfieldErrorCode } from '../public/core';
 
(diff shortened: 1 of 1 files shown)


arnog:
Thanks for the PR! Great fix...

igoracmelo:
@arnog @EaswarRaju 
For the record, this fixed an even worse bug:

on Safari webview the **fonts weren't even being loaded** because of this condition, and I spent weeks trying to fix it in my app.
Upgrading for 0.76.0 fixed it for me.

image

Issue #1523, opened 2022-06-30
[Regression] Body accumulates padding on virtual keyboard toggling


sbondaryev:
Version 0.76.1
The keyboard container accumulates `padding-bottom` while keyboard panel switching
![padding](https://user-images.githubusercontent.com/2843147/176785234-444b589d-0cf2-4997-8cc4-b5ae21473a25.gif)

Cosed by `showVirtualKeyboard` that sets `keyboard.visible = false;`and makes it possible to run the next code twice
without reseting the style of the container

https://github.com/arnog/mathlive/blob/f6dda8f1b7bf4260268def15468433e5af1b4aaf/src/editor/virtual-keyboard-commands.ts#L219-L227

Possible solution is to chek that `--keyboard-height` is set only once
```javascript
    const padding =
      keyboard.options.virtualKeyboardContainer.style.paddingBottom;

    if (!/--keyboard-height/.test(padding)) {
      keyboard.originalContainerBottomPadding = padding;
      keyboard.options.virtualKeyboardContainer.style.paddingBottom = padding
        ? `calc(${padding} + var(--keyboard-height, 276px) - 1px)`
        : 'calc(var(--keyboard-height, 276px) - 1px)';
    }
```

sbondaryev:
But I think it is better to rewrite toggle in terms of show/hide and chek for the part that modify the padding of the container

I've created a PR https://github.com/arnog/mathlive/pull/1524

PR #1524, opened 2022-06-30
rewrite toggleVirtualKeyboard in terms of show/hide VirtualKeyboard 

sbondaryev:
and fix https://github.com/arnog/mathlive/issues/1523

I added a check for the container padding modification in `showVirtualKeyborad`

https://github.com/arnog/mathlive/pull/1524/files#diff-4326e5672802bba8b043a8c8d6ce9ff0c70475dae04afc6072502c21e765dbccR202-R211

I don't know if we should execute the rest of the code in case when the keyboard is already visible, so I preserve the current funcionality


arnog (approved):
Looks good. Thank for the PR

PR #1593, opened 2022-08-23
Scale fix for living in a transform-scaled environment

gunnarmein-ts:
This is a tiny bit of code to make sure the selection is properly painted in an environment where a parent of MathfieldElement might have been scaled. In such environments, clientX/clientY/getBoundingClientRect() data cannot be used to construct coordinates to place new elements. Althought we cannot use getComputedStyle() to find the parent transform, we can compare sizes:

The fix calculates the ratio of the "supposed" width of the mathfield (from getComputedStyle) compared to the actual (client) width, and calculates the applied scaleFactor. Later, selection box coordinates are divided by the scaleFactor for both the collapsed and uncollapsed cases.

### DIFF
--- a/src/editor-mathfield/pointer-input.ts
+++ b/src/editor-mathfield/pointer-input.ts
@@ -126,8 +126,8 @@ export function onPointerDown(
     }
 
     // ts test
-    actualAnchor /= 2;
-    focus /= 2;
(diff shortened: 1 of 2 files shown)


arnog:
Clever way to find the scale factor! Thanks for the PR.

PR #1599, opened 2022-09-01
Enhance delete & insert delims in leftright atom

zeyad-ahmad-aql:
- keep leftright atom if one of its delimiters is removed.
- only hoist body if both delimiters are removed.


arnog:
Looks good, with one caveat: it's not necessary to insert a `\mleft...\mright` after functions anymore. The `\left...\right` command are now smart enough to do the right thing.

PR #1606, opened 2022-09-03
fix for issue 1605

gunnarmein-ts:
See issue for problem description. The problem is that nearestAtomFromPointRecursive initializes the result incorrectly, and children are not considered as closer candidates if the group is a "root". The fix is to broaden the criterion for special initialization from "group" only to "group" or "root".


arnog:
LGTM. Thanks for the fix.

PR #1607, opened 2022-09-03
modification for drag and drop

gunnarmein-ts:
This is the minimal change that will allow me to implement drag-and-drop on top of your control - if you click in the selection (pointerdown), do nothing. I don't expect that you will accept the change like this, but I wanted to show you what works. Perhaps this behavior could be set by a new option? Your options mechanism looks is complicated (to me), that's why I did not prototype one. Thanks!


gunnarmein-ts:
Sorry, when I started explaining this to you in email I realized I don't actually need it. Closing.

Issue #1631, opened 2022-09-20
Clicking and positioning the caret, revisited


gunnarmein-ts:
## Description

This is technically a re-opening of #1605. Clicking will not always position the caret close to the appropriate atom. I supplied the original fix, which set the distance of "root" atoms initially to infinity. There are apparently more cases, though - sometimes the container is neither "group" nor "root", and the distance is set to 0. I was not quite clear on how many container types there are when I wrote the fix. I have a better, simple fix that I will submit - this one never initializes distance to 0, and simply fills in the atom in cases where no child was found adequate *or no children exist*

## Steps to Reproduce

enter "x*(3+4+5)"
click on mathfield close to the "4". 
wiggle mouse. 

### Actual Behavior

sometimes, caret / selection end ends up on opening or closing parenthesis.

### Expected Behavior

Caret / selection should stay close to the "4"

### Environment

current master branch
MacOS Chrome

PR #1632, opened 2022-09-20
fix for #1631

gunnarmein-ts:


arnog:
Thank you for the PR! I think this is a good improvement...

PR #1634, opened 2022-09-22
options for offsetFromPoint, and expose nearestBoxFromPoint

gunnarmein-ts:


arnog:
Thanks for the PR! Looks good, but I think I would have a `hitboxFromOffset()` function instead: this would be a bit more flexible since you wouldn't need to have a point to get access to the hitbox. What do you think?

gunnarmein-ts:
I am highlighting the element the user is trying to drop on, and the offset gives me only the left or right of it and I am afraid the user will sometimes not understand why.

That said, I could explicitly ask for the left offset (with bias -1) from the point, then use hitBoxFromOffset() to get the same thing? Let me know whether you think that is correct.

From: Arno Gourdol ***@***.***>
Reply-To: arnog/mathlive ***@***.***>
Date: Thursday, September 22, 2022 at 11:35 AM
To: arnog/mathlive ***@***.***>
Cc: Gunnar Mein ***@***.***>, Author ***@***.***>
Subject: Re: [arnog/mathlive] options for offsetFromPoint, and expose nearestBoxFromPoint (PR #1634)


Thanks for the PR! Looks good, but I think I would have a hitboxFromOffset() function instead: this would be a bit more flexible since you wouldn't need to have a point to get access to the hitbox. What do you think?

arnog:
> That said, I could explicitly ask for the left offset (with bias -1) from the point, then use hitBoxFromOffset() to get the same thing? Let me know whether you think that is correct.

Yes, that's what I was thinking. It should still satisfy your use case but gives more options in case someone did not have a point position to start from...

gunnarmein-ts:
I will give it a shot and update when done.

From: Arno Gourdol ***@***.***>
Reply-To: arnog/mathlive ***@***.***>
Date: Thursday, September 22, 2022 at 1:37 PM
To: arnog/mathlive ***@***.***>
Cc: Gunnar Mein ***@***.***>, Author ***@***.***>
Subject: Re: [arnog/mathlive] options for offsetFromPoint, and expose nearestBoxFromPoint (PR #1634)


That said, I could explicitly ask for the left offset (with bias -1) from the point, then use hitBoxFromOffset() to get the same thing? Let me know whether you think that is correct.

Yes, that's what I was thinking. It should still satisfy your use case but gives more options in case someone did not have a point position to start from...

gunnarmein-ts:
Done.

arnog (approved):

PR #1639, opened 2022-09-25
Better vertical navigation

manstie:
I made it measure pixels and compare to the (base) elements of the upper or lower matrix cell. The way I get the DOM elements feels hacky.

LaTeX I used to test:
```
\begin{align}b+c+d+e & =f \\ a+\sqrt{a^2+b^2}+b & =g\end{align}
```

You can go from in the middle of the sqrt to the same (closest) position above. It doesn't work in reverse - not sure if that is desired?

Also this logic could probably be applied to moving between fraction upper and lower, but I haven't spent the time to look for where the code is for that.

I also made sure it works when the element is rotated by CSS.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -36,5 +36,6 @@
   "eslint.alwaysShowStatus": true,
   "eslint.format.enable": true,
   "eslint.lintTask.enable": true,
-  "cSpell.language": "en,en-US"
+  "cSpell.language": "en,en-US",
+  "editor.tabSize": 2
 }
--- a/src/core/atom-class.ts
+++ b/src/core/atom-class.ts
@@ -890,7 +890,7 @@ export class Atom {
   }
 
   get hasChildren(): boolean {
-    return this._branches && this.children.length > 0;
+    return Boolean(this._branches && this.children.length > 0);
   }
 
   get firstChild(): Atom {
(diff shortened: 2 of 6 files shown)


arnog (changes requested):
- could you use the existing `getAtomBounds()` function instead of `getLocalDOMRect()`?
- `getHTMLElement()` doesn't belong in the model. The model generally doesn't know about layout. This could be a method on the `Atom` class, except I don't think you can access the mathfield from it, so a method on `MathfieldPrivate` would work.
- Could you describe how to test this change (i.e. what the current implementation does vs your proposed changes?

manstie:
`getAtomBounds()` would definitely work for a general implementation and I would happily change it to that for the main branch (I would keep using `getLocalDOMRect()` for the kedyou branch as we require css rotation support).

This implementation goes off how the cursor should move visually compared to the previous index implementation.

The previous implementation only went by the indexes of the base atoms in the matrix cells:
image

The new implementation moves the caret based on x coordinate distance (without jumping into children):
image

But you can jump out of children to the closest spot above:
image

arnog:
OK, got it. Yes, very nice improvement.

For `getAtomBounds()`, I agree that CSS rotation support is nice to have. How about moving `getLocalDOMRect()` to `utils.ts` and replacing the `getBoundingClientRect()` in `getNodeBounds()` with `getLocalDOMRect()`?

manstie:
I think there is room for both implementations - `getLocalDOMRect()` is less performant, and it also loses decimal precision, so it should only be used for relative measurements such as this.

https://www.measurethat.net/Benchmarks/Show/21230/0/getboundingclientrect-vs-offset-when-taking-into-accoun

PR #1649, opened 2022-10-01
feat: make vertical navigation of fractions and matrices the same

manstie:
Same as #1639 but works with fractions (and anything else with an above/below?)

I hadn't considered the other else if branch of vertical movement - but conveniently `branch` is also an array of atoms so I can apply the same logic to an upper/lower branch as a matrix cell.

Tested with this LaTeX:
```
\begin{aligned}\frac{100000000000000000}{\sqrt{1x2y3z}+x+y+z} & =a \\ aaaaaaaaaaaaaaaaa & =b \\ bbbbbbbbbbbbbbb & =c\end{aligned}
```

### DIFF
--- a/src/editor-model/commands.ts
+++ b/src/editor-model/commands.ts
@@ -419,6 +419,47 @@ function getClosestAtomToXPosition(
   return search[i - 1];
 }
 
+function moveToClosestAtomVertically(
+  model: ModelPrivate,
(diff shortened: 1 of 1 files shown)


arnog:
Fantastic! Thanks for this PR.

arnog (approved):

PR #1666, opened 2022-10-09
feat: give empty environments placeholders, limit environment columns

manstie:
Made it so when you type in `\begin{some environment}` you get placeholders instead of empty brackets or empty space which lets you better know where you are inserting.

Also limited the columns when adding/removing using hotkeys based on `minColumns` and number of `align` elements in `colFormat`.

### DIFF
--- a/src/core-atoms/array.ts
+++ b/src/core-atoms/array.ts
@@ -250,7 +250,7 @@ export class ArrayAtom extends Atom {
     this.colSeparationType = options.colSeparationType;
     // Default \arraystretch from lttab.dtx
     this.arraystretch = options.arraystretch ?? 1.0;
-    this.minColumns = options.minColumns || 1;
+    this.minColumns = options.minColumns ?? 1;
   }
 
   static fromJson(json: AtomJson, context: GlobalContext): ArrayAtom {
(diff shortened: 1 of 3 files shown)


src/core-atoms/array.ts
```diff
@@ -248,6 +250,7 @@ export class ArrayAtom extends Atom {
     this.colSeparationType = options.colSeparationType;
     // Default \arraystretch from lttab.dtx
     this.arraystretch = options.arraystretch ?? 1.0;
+    this.minColumns = options.minColumns || 1;
```
arnog, line 253: Use `??` instead of `||` for default values

src/editor-model/array.ts
```diff
@@ -137,6 +137,11 @@ function addCell(
         break;
 
       case 'after column':
+        if (
+          arrayAtom.colFormat.filter((col) => Boolean(col['align'])).length <=
+          arrayAtom.colCount
+        )
+          return;
```
arnog, line 144: Could do a `model.announce('plonk')` before `return`. It's used to indicate to acknowledge the command, and to indicate the command is not currently applicable. It plays a "plonk" sound when the sound file is available.

src/editor-model/array.ts
```diff
@@ -149,6 +154,11 @@ function addCell(
         break;
 
       case 'before column':
+        if (
+          arrayAtom.colFormat.filter((col) => Boolean(col['align'])).length <=
+          arrayAtom.colCount
+        )
+          return;
```
arnog, line 161: Same as above re: `plonk`.

arnog (approved):
Looks good to me. Minor tweaks suggested

arnog (approved):

PR #1676, opened 2022-10-16
feat: allow pasting of multiple columns in environments

manstie:
There is a preexisting bug where if you highlight the entire innards of the environment and paste it will paste outside - I'm not sure how to address this...

https://user-images.githubusercontent.com/12475567/196035430-52cc342b-7298-4939-b95f-994c21faf064.mp4

Also not sure how to handle multiple rows... at least it's a start.

Also fixed an issue when copying from multiple columns to outside an environment - it included the `first` elements which makes things weird when they aren't in the first position.

### DIFF
--- a/src/core-atoms/array.ts
+++ b/src/core-atoms/array.ts
@@ -620,9 +620,27 @@ export class ArrayAtom extends Atom {
     return this.array[row][col];
   }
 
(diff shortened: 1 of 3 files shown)


arnog:
LGTM. Thanks!

PR #1785, opened 2023-01-14
fix: remove extra complete call

### LINKED ISSUE #1783: Escape character doesn't function when a SharedVirtualKeyboard is visible
## Description

As the title describes. When you use `makeSharedVirtualKeyboard` the escape character `\` doesn't function properly.

https://user-images.githubusercontent.com/12475567/211192773-da821faf-669d-4480-b7d0-215efe7328f4.mp4

## Steps to Reproduce

1. Replace the script with the following code in `examples/latex/index.html`

```
    <script type="module">
      import { makeSharedVirtualKeyboard } from '/dist/mathlive.mjs';

      const keyboard = makeSharedVirtualKeyboard();
      keyboard.buildAndAttachElement();

      const mf = document.getElementById('mf');
      mf.setOptions({
        smartMode: true,
        virtualKeyboardMode: 'manual',
      });
      mf.addEventListener('input', (evt) => {
        document.getElementById('latex').value = mf.getValue();
      });
      document.getElementById('latex').addEventListener('input', (ev) => {
        mf.setValue(ev.target.value);
      });
    </script>
```

2. npm start
3. Navigate to `/examples/latex`
4. Open the virtual keyboard
5. Use the escape character `\` or press `Esc`

### Actual Behavior

If the keyboard shown is one made by `makeSharedVirtualKeyboard` it causes the escape character `\` to fail.

### Expected Behavior

The escape character should work regardless of what type of keyboard is visible.

### Environment

**MathLive version**
 0.86.0

**Operating System**
Windows 10 Build 19044.2364

**Browser**
Firefox 108.0.1
Microsoft Edge 108.0.1462.76
Google Chrome 108.0.5359.125

manstie:
Closes #1783 

I couldn't find any references to `lower-command`, `upper-command`, `symbols-command` in the repo, and `complete` already gets executed in `mathfield-private.ts:1172`, so I'm not sure if this code is needed and it fixes the issue.

The issue being that because it's a remote keyboard anything going through `executeCommand` is delayed by 1 frame because it's handled by event listeners? So it was completing the inserted `\` one frame later.

### DIFF
--- a/src/editor/virtual-keyboard-commands.ts
+++ b/src/editor/virtual-keyboard-commands.ts
@@ -28,15 +28,6 @@ export function switchKeyboardLayer(
   keyboard: VirtualKeyboard,
   layer: string | null
 ): boolean {
-  if (
-    layer !== 'lower-command' &&
-    layer !== 'upper-command' &&
-    layer !== 'symbols-command'
-  ) {
-    // If we switch to a non-command keyboard layer, first exit command mode.
-    keyboard.executeCommand('complete');
-  }
-
   showVirtualKeyboard(keyboard);
   // If the alternate keys panel was visible, hide it
   hideAlternateKeys();


arnog:
Yes, this is OK to remove. The 'lower-command', 'upper-command' and 'symbols-command' layers have been renamed 'lower-latex', 'upper-latex' and 'symbols'-latex', but they no longer need this special 'complete' to work correctly, so it's safe to remove.

arnog:
Actually, I take that back. That call is still needed to exit the 'latex' mode when switching to a non-latex layer. I'll find some other way of dealing with this case.

arnog:
I've decided to drop the 'latex' virtual keyboard, so it's all good now.

PR #1792, opened 2023-01-15
fix: use cursor pointer on just the buttons of the virtual keyboard

manstie:
The entire keyboard had `cursor: pointer` when it should just be the buttons you can click.


arnog:
Great fix. Thanks!

PR #1806, opened 2023-01-30
Add matrix support for latex to speakable text

androettop:
This PR adds support for convert matrices in latex to speakable text. It also adds a new output in the Mathlive Smoke test page to test this feature.

### DIFF
--- a/src/editor/atom-to-speakable-text.ts
+++ b/src/editor/atom-to-speakable-text.ts
@@ -6,6 +6,7 @@ import { toMathML } from '../addons/math-ml';
 import { LeftRightAtom } from '../core-atoms/leftright';
 import { isArray } from '../common/types';
 import { osPlatform } from '../common/capabilities';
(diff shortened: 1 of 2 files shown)


NSoiffer:
I'm surprised matrix was missing. If you are checking for `(...)`, shouldn't you also check for `[...]` because square brackets are also used?

If matrix is missing, I'm guessing determinant is also missing, so adding something for `|....|` would make sense -- as it also needs to iterate over all the children (hence code reuse).

androettop:
@NSoiffer You are right! I have just added support for all the matrix types listed in the documentation.

- array
- matrix
- pmatrix
- bmatrix
- Bmatrix
- vmatrix
- Vmatrix
- matrix*
- smallmatrix

arnog:
@NSoiffer I don't think that matrix support was implemented in MathLive when you worked on the speech support.

@androettop thank you for this PR!

PR #1839, opened 2023-02-17
fix for ctrl x/c/v

gunnarmein-ts:
trivial addition of inputtype="none" on textarea seems to successfully disable OS virtual keyboard, but otherwise stay functional.


arnog:
The attribute `inputmode` was already set to `"none"`. So, always having the `<textarea>` would work with this setting? Interesting. Let's try it!

arnog:
@gunnarmein-ts I'm working on an alternate fix.

Issue #1838, opened 2023-02-17
Alternative approach to "fill in the blank"


WildYellowfin:
I'd like to suggest an alternative approach to the "fill in the blank" feature. The current placeholder solution uses hacky getBoundingClientRect() stuff that doesn't seem stable. I also don't think putting nested math fields inside a read-only is a solid approach.

Instead I would make a new \prompt[id?][style?] command and a prompt mode. In prompt mode, the typical focus UI would be disabled, and selection would be controlled to give the illusion that the non-prompt parts are read only. 

\prompts would basically be boxes, and inherit border styling from the math field. They would also inherit the :focus-within styles. 

Prompts can be controlled by RegExing the global latex (I feel like there's a better solution, but don't know the library well).

I'm fairly new here, so I want to know how feasible this is. Given a few pointers I'd be happy to build this.

arnog:
Thanks for your interest and sharing this idea.

That's something interesting and worth thinking about a bit more.

Could you expand a bit on what issues you see with the current implementation of `\placeholder{}`? What problems would your approach solve?

It seems that you're arguing for a different implementation of `\placeholder{}` rather than a new command. I'm not sure how you would explain to a user of a math-field whether to use `\placeholder{}` or `\prompt{}`. Seems they would do the same thing but with a different implementation?

I don't think there's anything unsolvable, but there are some cases that are worth thinking about in advance. I would put them in three categories:
- keyboard input: with this approach, the selection of a mathfield could span a read-only portion and an editable portion. For that matter the selection could start in a read-only portion, completely include an editable portion, span some more read-only portion, and end in the middle of an editable portion. In that case, what happens if the delete key is pressed? Or another key, for that matter. What happens when the right arrow key is pressed at the end of an editable portion? What does command-A do?
- pointer input: you can drag a pointer to select a portion of mathfield. What happens when your cross an editable and non-editable portion? If you are currently in an editable portion, what happens if you click on a non-editable one?
- API behavior: what would `mf.value` be when including a read-only and editable portion? Can you query the value of the portion that's editable? Can you get notified (event) when an editable portion is modified? What happens if you set the value of the mathfield, or perform an `mf.insert()`?

> Prompts can be controlled by RegExing the global latex (I feel like there's a better solution, but don't know the library well).

I'm not sure what you mean by that.

[...]

PR #1857, opened 2023-02-28
Fill in the blank rework

### LINKED ISSUE #1842: [bug] Move-out event not called when caret in a matrix
I found something like a bug.
When the caret is at the bottom of a matrix like 
スクリーンショット 2023-02-20 13 47 09
typing "ArrowDown" does not call the "move-out" event although the caret does not move. 
I use "mathlive": "^0.86.0" and MacOS 13.2.1（22D68）.

WildYellowfin:
#1838 
### Summary
- A "blank" is no longer a separate math field, but instead a styled latex element that **looks**  like a mathfield
- The latex syntax is `\placeholder[id][correctness?]{body}`. IDs are not enforced to be unique, but duplicate IDs will mess with behavior of related commands.
  - The old syntax `\placeholder[id][defaultValue?]{}` is still supported, but will be serialized from now on as `\placeholder[id]{defaultValue}`. If a `body` is provided, `defaultValue` is ignored.
  - `correctness` is read through this same argument space, but it is inferred to be a default value if the content is not `"correct"` or `"incorrect"`.
- Any ID'd prompt in a `readOnly` mathfield is editable, unless it has `correctness` set.

### Keyboard navigation:
- Left/right arrows + tab/shift tab move to next / previous editable prompt. Move-out fires if none
- Up/down arrows work in fractions, matrices etc.

### APIs
- `getPrompt(id): string`
- `getAllPrompts(): string[]`
- `setPromptContent(id, content)`
- `setPromptCorrectness(id, correctness: 'correct' | 'incorrect' | undefined)`

### Styling
- If not `readOnly`, prompts are greyed out
- In `readOnly`, prompts are styled like a mathfield.
- If `correctness` provided, box shadow is added. On `incorrect`, box will be "crossed out" 
- Correct / incorrect outline and shadow color can be set in `--correct-color` / `incorrect-color` (green/red by default)

Other:
- Spoken text supported ('start input, content, end input, correctness.')
- Fixed #1842 
- Fixed `enclose` serialization

https://user-images.githubusercontent.com/104928632/221940034-980f1e46-a59a-427e-8aa4-d5ae25229171.mov

Suggested documentation example:
\frac{\placeholder[0]{x+y}}{\placeholder[1]{\sqrt{2}}} \begin{pmatrix}
        \placeholder[2][incorrect]{} & \placeholder[3]{} & \placeholder[4]{} \\
        \placeholder[5]{} & \placeholder[6]{} & \placeholder[7][correct]{}
        \end{pmatrix}


arnog:
This looks really good. Thanks for this PR.

One small feedback: I'm wondering if it would be useful to add some other states for correctness. Currently, a prompt with a corrected property is not editable, but it seems that in some use cases it might be useful to have (i.e. so you can go back and modify your answer if it was wrong). Perhaps there could be four states: `"correct"`, `"incorrect"`, `"correct-locked"` and `"incorrect-locked"`? And maybe a fifth `"locked"`, although less clear if that would be useful?

Also, there are some conflicts with the head now (sorry, I checked-in a bunch of changes recently). Do you want to go through them or would you like me to?

WildYellowfin:
I guess I can see that being useful, it just hadn't occurred to me since its outside my use case. I might just make locked / unlocked a third argument, since I think the behavior can almost always be inferred just from correctness. 

As for conflicts, they all seem trivial enough. I'll see if I can commit some conflict free version tonight.

WildYellowfin:
Done! `[locked]` can now be added as a third argument, and defaults to `false`. I ended up removing `readOnly` from `GlobalContext` since with `locked` I don't think a prompt should always be non-editable in `readOnly`. 

There is also now a `setPromptLocked(id, locked)`. 

I'm happy to maintain this as needed, since its a pretty core part of the platform we're building. Feel free to @ me on any related issues.

arnog:
@WildYellowfin an issue has been reported. I'm not sure why you made the change, so I reverted it for now, but happy to discuss the proper way to handle what you were trying to do (see bb3d25e4c46af54715d556162579cdd25c96e712).

WildYellowfin:
@arnog ahh this wasn't an intended change. I was trying to debug the Webkit `inputType` issue and just forgot to remove it...

PR #1887, opened 2023-03-24
fix: firefox focus issues, fixes #1861

### LINKED ISSUE #1861: Firefox: No keyboard input on seemingly focused math-field
## Description
On Firefox (version 110, tested on Ubuntu 22.04 and macOs Ventura) If you have an empty math-field and click on the border of the field (somewhere in the middle) it's sometimes possible to get it to seemingly give focus to the math-field, but having no input when pressing the number keys on the keyboard, letters does usually not produce anything either, but sometimes pressing letters randomly will "unlock" it and now everything works (perhaps I'm randomly hitting a short cut for something, and that's what's unlocking it).
It's also possible to "unlock" it by clicking close to the blinking cursor. Using the virtual keyboard to enter anything also "unlocks" it.

When this happens the `keydown` event still triggers, but no `input` event (in keyboard.ts keyboardSink's 'keydown' and 'keypress' are triggered but not 'input')

I have reproduced this both on [the demo page](https://cortexjs.io/mathlive/demo) and on Mathlive Smoke Test. It happens on both the latest master (as of writing 04a23af4c74b43d8a1a59e5c7115f9292af24161) and on 0.89.4. But it does not happen on 0.89.2.

This does not happen on Safari or Chrome.

## Steps to Reproduce

1. Run firefox and open Mathlive Smoke Test
2. Clear out the math-field so it's empty.
3. Un-focus the field by clicking elsewhere.
4. Click on the border of the field somewhere in the middle length wise.
5. The field becomes focused, but won't let you input any numbers using the keyboard.

It doesn't always happen, but fairly often. The math-field does not have to be empty to trigger this, but if it isn't empty then just clicking on any of the maths will unlock it. 

In my environment it happens every time the math-field loads (it always starts out empty), while it only sometimes happens in Mathlive Smoke Test. I haven't yet figured out what the difference that's causing to always fail for me.

### Environment

**MathLive version**
0.89.4 - Affected
0.89.2 - Not affected

**Operating System** [macOS, Windows, iOS. Include the version]
Ubuntu 22.04
macOs Ventura 13.2.1

**Browser** 
Firefox 110 - Affected
Chrome 110 - Not affected
Safari - Not affected

mgreminger:
Sets `delegatesFocus` to `true` for shadow DOM, this fixes the Firefox focus issues. Just setting `delegatesFocus` to `true` breaks tab focusing. However, setting `tabindex` to always be `0` for `span.ML__keyboard-sink` allows the math field to receive tab focus again. I'm not sure if this change to `tabindex` for `span.ML__keyboard-sink` will have other side effects. However, tab focusing seems to working properly in the smoke test and in my app (tab focusing seems to be working with the prompts as well). 

I'm currently working on migrating the open source engineering calculation app [EngineeringPaper.xyz](https://github.com/mgreminger/EngineeringPaper.xyz) from [MathQuill to MathLive](https://github.com/mgreminger/EngineeringPaper.xyz/pull/167) and this is one of the blocker issues. Without addressing this issue, the app is unusable in Firefox. 

Many thanks for creating the excellent MathLive library. I'm excited to get my app moved over to MathLive to make it easier to add new functionality such as vectors and matrices.


mgreminger:
Also, forgot to mention that this change fixes the issue that `math-field` elements didn't previously emit `focusin` or `focusout` events. With this change, these events are now emitted, making the `math-field` behavior more similar to a `textarea` element.

mgreminger:
I am noticing one issue, with this change, `readOnly` math fields are receiving tab focus. I'll need to look at this a little more before it's ready, I should get a chance tomorrow.

arnog:
EngineeringPaper is a very cool project! Love it.

Let me know when you have figured out a combination of settings that make FireFox happy!

arnog:
BTW, read-only mathfields *should* receive focus. Same behavior as `<textarea>`. I think the idea is that you might want to copy their content, or at least know that there is a field there, but that it cannot be edited. Anyway, that's the spec.

Disabled mathfields on the other hand should *not* receive focus. 

The `test/mathfield-states` test case has some side-by-side comparison with textarea.

mgreminger:
Thanks for the info, yes that makes sense for `readonly` elements. I've tested the `mathfield-states` page and the tab focusing is not impacted by this pull request. Also, tested setting `tabindex="-1"` on the math-field element, and that takes the math field out of the tab order, as it should. 

I did notice on the `mathfield-states` page that setting `user-select: none` is trapping tab focus. However, this occurs with or without this pull request, so it's a separate issue. I'll open a separate GH issue for this one.

I did notice notice in testing that the `+` and `=` virtual keyboard buttons are both inserting `-` into the math field. I do have a staged commit that fixes this. Do you have a preference as to whether I should add the commit to this pull request or open a separate pull request for the keyboard fix?

arnog:
Keep the fix for +/- with this PR.

mgreminger:
Added the keyboard commit, this one should be good to go from my end. It's working really slick in Firefox now.

arnog (approved):

arnog:
Thanks!

Issue #1909, opened 2023-04-02
Virtual keyboard does not work


zxhd863943427:
## Description

After using vite and vue to package, there will be no response after the virtual keyboard in the project is pressed

## Steps to Reproduce

1. Directly download the running file of [unpkg](https://unpkg.com/mathlive@0.90.11/dist/mathlive.min.js) and pack it into the project
2. Render the virtual keyboard in the project
3. Press a key on the virtual keyboard

### Actual Behavior

virtual keyboard won't work

### Expected Behavior

The virtual keyboard takes effect

Works fine in 0.90.10

### Environment

**MathLive Version** : 0.90.10

**OS** : Windows10

**browser** : electron

zxhd863943427:
I have solved this problem, but it takes a little time to upload Pull requests.
I'm sorry that my ts is not very good, I checked and solved the problem directly through the packaged mathlive.js file.

arnog:
The `Node` class is no longer referenced, so this should no longer be a problem.

PR #1910, opened 2023-04-02
fix the bug of Virtual keyboard does not work in vite

zxhd863943427:
fix https://github.com/arnog/mathlive/issues/1909

### DIFF
--- a/src/virtual-keyboard/utils.ts
+++ b/src/virtual-keyboard/utils.ts
@@ -1208,7 +1208,7 @@ function isKeycapElement(el: Node): el is HTMLElement {
 export function parentKeycap(
   el: EventTarget | Node | null
 ): HTMLElement | undefined {
-  if (!el || !(el instanceof Node)) return undefined;
+  if (!el || !(el instanceof globalThis.Node)) return undefined;
 
   let node: Node | null = el as Node;
   while (node && !isKeycapElement(node)) node = node.parentNode;


arnog:
Thanks for reporting this and debugging the issue.

I'm not sure I understand the fix, though. The `Node` class should be a global variable, and so it should already be synonym with `window.Node` or `globalThis.Node`. I'm concerned that the fix may only be partial and that there may be some other issues lurking.

It does seem like something is overriding the `Node` class, somehow.

In your environment, what is `Node` equal too? That is, if you do `console.dir(Node)` in the console, what do you get?

arnog:
OK, that makes sense.

So it seems like the `Node` function in `Yallist` is what's causing the problem. The thing is, that function should not be imported in the scope of MathLive, so it shouldn't interfere with the DOM `Node` class.

Do you use a forked version of MathLive in which you import `Yallist`? How do you import `Yallist` and `MathLive` in your project?

If your project is in GitHub, I'd be happy to have a look at it.

zxhd863943427:
I'd be more than happy to tell you the address of the project, It is in:
https://github.com/zxhd863943427/siyuan-math-enhance

The way I import mathlive is as follows: 

https://github.com/zxhd863943427/siyuan-math-enhance/blob/a7c4ef00d87ef487092ac7bbd08ad0873e5c945d/src/main.ts#L10

As you can see, I am not using a conventional method to import mathlive, because this is a plugin written for SiYuan-Note, because mathlive installed with npm needs to export font files, and the current plugin system cannot support it other than main.js, so I had to import that way.

SiYuan-note is a local-first personal knowledge management system，it is in: https://github.com/siyuan-note/siyuan

It also has a community-supported playground at http://www.siyuanplay.com

arnog:
I've removed the dependency on the `Node` class, so I don't think this patch is needed any longer.

PR #1911, opened 2023-04-02
insertPrompt command and widen empty prompt, allow latex mode in prompt

WildYellowfin:
`insertPrompt` command inserts a prompt with random 6 char id or given `id`.

Prompts with no content were too small, so I widened them to `3 * fboxsep` (previously 2)


WildYellowfin:
Ok! Will commit this evening.

PR #1912, opened 2023-04-03
[draft] tests: add playwright end-to-end browser tests

mgreminger:
This pull request adds [Playwright](https://playwright.dev/) end-to-end browser tests. Playwright is able to test against all of the major browser engines (Chromium, Firefox, and Webkit). This initial version has the basic framework implemented and an initial basic browser test using the smoke test page. I've also added a Github action to run the test suite (unit, type, and playwright) for commits to pull requests and commits to the main branch. The build and test process is the same as before with the exception that the playwright browsers need to be installed the first time a project is setup (or anytime playwright is upgraded) using the `npx playwright install` command. I've added instructions to the contributing document with these instructions in addition to the basic build instructions for MathLive. 

Let me know if this is something of interest to add to MathLive and I'll implement some additional tests (virtual keyboard, space bar handling, tab focus, editing options, etc.). 

These tests will be helpful for catching issues with particular browsers and to prevent regressions as features are added. The tests will be fragile at times. For example, insignificant changes, such as white space changes, to latex results will cause tests to fail and require them to be updated. However, I think the benefits will outweigh the inconveniences.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,0 +1,29 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ main, master ]
+  pull_request:
+    branches: [ main, master ]
+jobs:
+  test:
+    timeout-minutes: 20
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 16
+    - name: Install dependencies
+      run: npm ci
+    - name: Build library
+      run: npm run build
+    - name: Install Playwright Browsers
+      run: npx playwright install --with-deps
+    - name: Run Playwright tests
+      run: npx playwright test
+    - uses: actions/upload-artifact@v3
+      if: always()
+      with:
+        name: playwright-report
+        path: playwright-report/
+        retention-days: 30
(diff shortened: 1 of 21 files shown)


arnog:
Thank you for this.

That sounds like it would be a great complement to the existing unit test (`npm t`) that focus only on the rendering.

mgreminger:
Sounds, good. I'll add a few more initial tests and let you know when it's ready to be considered for merging.

mgreminger:
Should be ready to merge now. The github action should just work once it's merge since I think that actions are enabled by default. Otherwise, you can go into the repo settings and enable actions . The github action script may need to tweaked some but I won't be able to tell until it runs. With the typescript fixes, all of the tests are passing on both Linux and MacOS

I just noticed that he latest keyboard change (9873f45) will probably cause the virtual keyboard test to fail. I'll update the virtual keyboard test today or tomorrow if you want to wait to merge this.

mgreminger:
This github branch on my fork has become corrupt and I'm unable to push the final commit (even browsing to the branch on my fork via the github site gives a 500 error). I'm working on resolving the issue...

mgreminger:
Closing this pull request, I was able to create a new branch with all of these changes and the pending changes, will create a new pull request shortly.

Issue #1914, opened 2023-04-04
virtual keyboard hides when moving focus off mathfield with mathVirtualKeyboardPolicy="manual"


albertchae:
## Description

Hi, I want to control visibility of the virtual keyboard manually (similar to described in this issue https://github.com/arnog/mathlive/issues/1345). I see that `virtualKeyboardMode` was recently renamed to `mathVirtualKeyboardPolicy`, but when I try setting this to `manual`, the keyboard still seems to hide when moving focus off the mathfield (it does prevent it from showing automatically when focusing on the mathfield though).

Here is a reproduction https://albertchae.github.io/mathlive-hide-when-focus-changed/. Source code is at https://github.com/albertchae/mathlive-hide-when-focus-changed. You'll see that when the page is loaded, the keyboard is shown, but if you click anywhere on the body, then the keyboard goes away. Is this intended behavior?

## Steps to Reproduce

1. Load https://albertchae.github.io/mathlive-hide-when-focus-changed/
2. In console, you can type `mathField.mathVirtualKeyboardPolicy` to verify it is set to `manual`
3. Click on the body (or the take focus button if tapping the body doesn't switch focus, like in iOS safari)

### Actual Behavior

The keyboard is hidden.

### Expected Behavior

The keyboard is not hidden, because we would control its visibility manually.

### Environment

**MathLive version** 0.90.11

I can reproduce this on iOS 16.3.1 with mobile safari and mobile chrome version 111.0.5563.101.

I also see this on macOS 13.2.1 with both Chrome 111.0.5563.146 and Firefox 111.0

### Recording with mobile chrome
https://user-images.githubusercontent.com/217050/229718798-104383ed-ed80-40bd-aef2-a02ef7cdf83c.MP4

### Recording with mobile safari
https://user-images.githubusercontent.com/217050/229718818-90517293-5a95-41ae-9d58-a041fc992811.MP4

arnog:
That's an interesting issue. 

The current behavior is as intended. It reflects my interpretation of the W3C virtual keyboard spec, which is a work in progress but something I was trying to follow (https://www.w3.org/TR/virtual-keyboard/#dfn-manual). On re-reading the spec, it actually does not speak clearly regarding this situation.

Since `mathVirtualKeyboardPolicy` is a property of a mathfield, it would suggest that it dictates what happens when the mathfield is focused. However, when no mathfield is focused, my interpretation was that since the virtual keyboard did not apply it should be hidden.

[...]

PR #1915, opened 2023-04-04
Don't hide the keyboard on focusout if mathVirtualKeyboardPolicy is manual

albertchae:
If #1914 is a bug, I think this is one way to fix it, although it's not great because `mathVirtualKeyboardPolicy` is not a global setting, so I have to save it off the most recently focused `mathField`.

Fixes https://github.com/arnog/mathlive/issues/1914


albertchae:
Never mind I tested this and I don't think it works. I erroneously thought https://github.com/arnog/mathlive/pull/1915/commits/9324c901c7c88957ef00f55036041c3be1e95f11 was working but it didn't hide the keyboard because `mf` was undefined, not because the logic was correct.

PR #1918, opened 2023-04-05
tests: set dev server to port 9029

mgreminger:
The default esbuild dev server port was varying between 8000 and 8001, was probably just look for the first available port starting at 8000. This pull request explicitly sets the port to 9029. Should not be a commonly used port. I've tested on Linux and MacOS, and it works on both.

Also, previously the github action was only running the playwright test. I've updated to run the whole test suite.


arnog:
Whoo-hoo! The tests work now. Thanks!

mgreminger:
No problem, glad to hear its working!

Issue #600, opened 2020-08-28
[FEATURE] Improved matrix editing UI


zwz:
I am quite new to mathlive.
And I find it not intuitive and user-friendly to do matrix editing, such as 
1. adding a column (I get it done by memorizing the keyboard shortcuts)
2. adding a new row for a multi-column matrix only gets one-column row
3. very hard to tell how many columns are there already when most columns are blank

So I suggest
1. add buttons of "new column/row (before/after)" and "remove current column/row" in the virtual keyboard
2. when a new row is added, it spans into multi-column row
3. each blank spot shows `⬚`(DOTTED SQUARE) as a placeholder. This will not only make it easy to tell the column-size, but also help user to locate and fill the spot to make a sparse matrix on a mobile device.

stefnotch:
I agree that the current handling of matrices could use some improvements.

For what it's worth, here is how some other editors handle this
**Swiftcalcs** (which I think uses a modified version of MathQuill)
- Add row `,`
- Add column `;`
- Shows placeholders

**Mathcha** (uses a really good, custom, closed source math editor)
- Add row and add column `Enter` and then choosing
![image](https://user-images.githubusercontent.com/10220080/91549765-b94c4080-e927-11ea-80c8-3f03ae1b9632.png)
- Shows placeholders

**Mathcad** (expensive desktop application)
- Add row `Shift`+`Space`
- Add column `Shift`+`Enter`
- Shows placeholders

NSoiffer:
Adding more systems:
**Mathematica**
*  add column after: ctrl+,
*   add row after: ctrl+enter/return (windows/mac)

Also menu entries for them along with a way to span and split them.

**Word**
Not really any keyboard shortcuts (tab adds a row at the bottom in some cases).
However, it has several menu entries and ribbon entries to add one or more rows/cols above and below current selection.

Options for merging and splitting

arnog:
To clarify, MathLive by default uses the same shortcuts as Mathematica (Ctrl+, and return to add a column/row). I don't think this issue is about what the shortcuts should be or that MathLive is lacking shortcuts for basic matrix editing.

Note also that it is possible to define a custom virtual keyboard with commands to add/remove columns/rows. If someone wants to submit one, I'd be happy to review it.

The point about having a placeholder visible for empty cells is a good one.

[...]

PR #1926, opened 2023-04-11
Matrix Controller, styles API

WildYellowfin:
This PR introduces a matrix controller:

https://user-images.githubusercontent.com/104928632/231286486-ebf48fe9-a84a-4585-8491-4413870d9080.mov

This panel can be resized using the `--environment-panel-height` setting. Aspect ratio is 2:1 so no width is necessary.
The left hand side of the controller is for moving around and manipulating arrays. I plan to make a `moveCell(direction)` command so that the arrows always move selection to the appropriate cell. 

In building this I needed to better type the `Environment`, so there are now `MatrixEnvironment`, `CasesEnvironment` and `TabularEnvironment` types, as well as type guards. These distinctions govern what shows up in the env panel when it comes to delimiters.

This PR also introduces a "selectionStyle" at `window.mathVirtualKeyboard.style`. This returns the intersection of styles for all selected atoms. An example use case (mine :) is displaying a rich text editor toolbar with styles information about the selection.


arnog:
I'm attempting a merge. I assume that 

in `get selectionStyle(): Style`
> `for (const [key, value] of Object.entries(style)`
should be 
> `for (const [key, value] of Object.entries(a.style)`

arnog:
The merge seems to work...

FYI, now typing `(` `alt/option`+`Tab` creates a `pmatrix`.

WildYellowfin:
Closes #600

PR #1927, opened 2023-04-12
Expose isSelectionEditable, add stripPromptContent, rework jot in arrays, fix environment popover

WildYellowfin:
1. In order to replicate the old `onfocus` keyboard show, developers need to tell whether a read-only mathfield's selection is within an editable prompt. So I added the existing `isSelectionEditable` get to the public mathfield.
2. Added a `stripPromptContent` method, which removes the content of all prompts matching a filter and returns those contents in a map. 

As an example use case of `stripPromptContent`, at [Perplex](https://perplex.org) we have a creator (ie. teacher) view where answers are typed in prompts, and we want to 'clean' these prompts before showing them to students. We had a lot of trouble doing so with existing methods (extra events, weird sync bugs requiring  too many timeouts...).


WildYellowfin:
#1930

WildYellowfin:
@arnog this is failing because I added 'jot' to the array rendering. Do you want to pull this now? I've [been having to add](https://perplex.org/solve/yCi32IxC) empty lines for spacing in environments...

arnog:
What do you mean by "added 'jot'"? Did you change the default value of the jot register? Also, what's the LaTeX for this https://perplex.org/solve/yCi32IxC ? Did you use `\\[]`?

WildYellowfin:
I noticed that `jot` was being totally ignored, so I added a line in the array render to take it into account. That latex is an `align` environment. The last two rows are the default spacing. If you’re seeing a `\[` it’s because I copied that latex from another problem I posted, but in any case I fixed it now.

arnog:
Ah, I see. Yes, using the `jot` register is the right approach. The `jot` property should be removed from the `ArrayAtom` (it dates back from before there was support for registers). Also, `jot` is not applied to `*matrix` and `array` environments.

WildYellowfin:
Is it applied to any matrices currently? I set it and saw no effect, and couldn’t find it being used anywhere. Do you mean removing the `arrayStretch` (I think it’s called)?

arnog:
`jot` is used in the definition of the `align` family of environments (seen `core-definitions/environments.ts:363`. It should be applied (see `core-atoms/array.ts:419`: `if (this.jot !== undefined) depth += this.jot`), but it's hard-coded to 0.3em, not respecting the jot register. Your implementation using the jot register is more correct. The `jot` property from `ArrayAtom` should just be dropped, although there should be a check to make sure the `jot` register is not applied to `matrix`-family and `array` environments. One way to do that I suppose would be to keep the `jot` property, and set it from the `jot` register at atom creation time.

I believe that `arrayStretch` is currently not supported, although really, it should.

WildYellowfin:
Ok! I think this latest commit does exactly what you say. For reference:
Asset 1

WildYellowfin:
Also, feel free to squash this - too many silly commits.

Issue #1845, opened 2023-02-23
When input abs value, the inner ")" will replace the last "|"


G4ngL:
## Description

When input abs value, the inner ")" will replace the last "|"

## Steps to Reproduce

1. enter "|" or click "abc" on the virtual board 
2. enter "()"

### Actual Behavior

expect to be "|(something)|"

### Expected Behavior

actual to be "|(something)"
missing the last "|"

### Environment

MacOS, Edge 110

![report-bug](https://user-images.githubusercontent.com/122529526/220834268-40553cab-938a-4563-9bc1-cb2b5389c4d4.gif)

Issue #1742, opened 2022-12-04
Fixing parens manually triggers malformed MathJSON


sritchie:
## Description

(@jefffriesen found this while playing with https://github.com/mentat-collective/mathlive.cljs/issues/11.)

to repro:

- type an expression like `1+cos()y`
- decide you want to pull `y` into the parens
- delete the right paren and retype it to the right of `y`

The `Mathfield` eagerly inserts a multiplication dot and parses the input as

<img width="186" alt="image" src="https://user-images.githubusercontent.com/69635/205501196-90bc6670-a6b8-4b84-8f0f-7fca331c36ad.png">

Full interaction:

![2022-12-04 08 48 41](https://user-images.githubusercontent.com/69635/205500976-6a0e8825-4abd-49a5-a62d-16c107ec6d41.gif)

MathJSON:

```
[
  "Sequence",
  [
    "Add",
    1,
    [
      "Sequence",
      [
        "Cos",
        [
          "Error",
          [
            "ErrorCode",
            "'expected-close-delimiter'",
            "\\right)"
          ],
          ["Latex", "'('"]
        ]
      ],
      [
        "Error",
        ["ErrorCode", "'expected-open-delimiter'", "\\left"],
        ["Latex", "'\\right.'"]
      ],
      "y"
    ]
  ],
  [
    "Error",
    ["ErrorCode", "'expected-open-delimiter'", "("],
    ["Latex", "')'"]
  ]
]
```

arnog:
I think that's a duplicate of #1656

gerryfletch:
I'm noticing this quite frequently with parens, especially because if I type `(` it'll create the corresponding `)` for me, but pressing backspace to write into it creates the `.`

arnog:
Fixed now.

Issue #1691, opened 2022-10-25
Inserting right paren first causes unbalanced parens 


shidel-dev:
## Description
See screen recording

https://user-images.githubusercontent.com/5149882/197809936-410d8c95-384e-4df3-97a2-1c0a6c8d5321.mov

## Steps to Reproduce
1. type number
2. type right paren
3. use arrow keys to move before number
4. enter left paren

### Actual Behavior

unbalanced parens

### Expected Behavior

balanced parens (an extra right paren should not be created as it is already in a balanced state)

Issue #1375, opened 2022-03-06
Input of curly brace '{' breaks things


manstie:
## Description

https://user-images.githubusercontent.com/12475567/156916017-4a0d7ae1-94d3-4f9c-af8c-3c0213cebd69.mp4

## Steps to Reproduce

Inputs:
`{`, `a`, `}`, `<right arrow>`, `<right arrow>`, `<backspace>`, `<backspace>`, `<backspace>`, `<delete>`, `{`, `a`, `}`, Click in the input field

### Actual Behavior

1. Input `{` -> Assertion failed `atom-class.ts:715:12`
2. Input `a` -> ok
3. Input `}` -> Assertion failed `atom-class.ts:743:12`
4. Input `<right arrow>` -> you get two cursors, one before and one after the `}`
5. Input `<right arrow>` -> cursor at right of `a`
6. Input `3x <backspace>` -> Displays `\lbrace` in the equation area and `\backslash` in the LaTeX area, has a cursor at the beginning and end of the line
7. Input `<delete>` -> removes the whole line
8. Input `{` ->
```
Uncaught TypeError: child.parent is undefined
    extractAtoms model-private.ts:339
    insertSmartFence mode-editor-math.ts:613
    insert mode-editor-math.ts:114
    insert mode-editor.ts:122
    onKeystroke keyboard-input.ts:299
    deferNotifications model-private.ts:618
    onKeystroke keyboard-input.ts:295
    keystroke mathfield-private.ts:439
    delegateKeyboardEvents keyboard.ts:246
    delegateKeyboardEvents keyboard.ts:229
    MathfieldPrivate mathfield-private.ts:407
    connectedCallback mathfield-element.ts:1083
    <anonymous> mathfield-element.ts:1755
```
9. Input `a`, `}` -> ok
10. Click -> Assertion failed `atom-class.ts:869:21` and
```
Uncaught TypeError: this.parent is undefined
    get rightSibling atom-class.ts:869
    onPointerDown pointer-input.ts:196
    handleEvent mathfield-private.ts:758
    on utils.ts:33
    MathfieldPrivate mathfield-private.ts:454
    connectedCallback mathfield-element.ts:1083
    <anonymous> mathfield-element.ts:1755
```

### Expected Behavior

1. Input `{` -> No error, LaTeX output should be `\lbrace`
2. Input `a` -> ok
3. Input `}` -> displayed equation should be `{a}`
4. Input `3x <backspace>` -> input area should be empty
[...]

PR #1932, opened 2023-04-17
Delimiter editing fixes

### LINKED ISSUE #1656: FEATURE: deleting a `\right` command should remove the matching `\left`
It is possible in some scenarios to end up with formulas that appear visually correct, but have mismatched `\left...\right` commands.

For example: type `(w-x` `DEL`, `\times10^6`, then go back after the `x` to insert a `)`.

The resulting formula is:
`\left(w-x)\times10^6\right.` which is visually identical to 
`\left(w-x\right)\times10^6` or 
`(w-x)\times10^6\`

Instead, when  explicitly deleting (forward delete or backspace) what visually appears as a `\right` command (i.e. when at the boundary of a `LeftRightAtom`), the `LeftRightAtom` should be demoted and the `\left(` replaced with a `(`.

mgreminger:
Fixes #1656 and #1742 (#1742 is a duplicate of #1656). This pull request also fixes a regression of #1691 (the #1691 regression occurred since 0.91.2, I didn't determine exactly which commit introduced the regression). 

Additionally, fixes issue where deleting left delimiter and retyping left delimiter did not trigger an input event.

End-to-end tests have been added for these fixes and the previous delimiter issues (#1691, #1375,  and #1845) to make sure regressions were not introduced with these changes and to make sure any regressions on these issues will be detected in the future.

### DIFF
--- a/src/editor-mathfield/mode-editor-math.ts
+++ b/src/editor-mathfield/mode-editor-math.ts
@@ -642,7 +642,7 @@ export function insertSmartFence(
     ) {
       parent.leftDelim = fence;
       parent.isDirty = true;
-      contentDidChange(model, { data: fence, inputType: 'insertText'});
+      contentDidChange(model, { data: fence, inputType: 'insertText' });
       return true;
     }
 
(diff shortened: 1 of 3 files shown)


mgreminger:
I added some additional fixes that handle cases where the left delimiter is deleted and replace either inside or outside of the LeftRight atom. I based it on how this is currently handled for the right delimiter case. I've added tests for both the left delimiter and right delimiter cases.

arnog:
Thank you!

Issue #1545, opened 2022-07-16
Mathfield lose focus on nav-tab changing


sbondaryev:
## Description
Mathfield lose the focus while one switches between the tabs of the navigator (or virtual screens on mac)

![focus](https://user-images.githubusercontent.com/2843147/179347096-e2609d4b-1f1e-4002-9259-07773e684be8.gif)

## Steps to Reproduce

1. Go to the demo page 
2. Open two different tabs in the navigator
3. Focus the mathfield in one tab
4. Switch between the tabs

### Actual Behavior

Mathfield lose the focus

### Expected Behavior

Mathfield should stay focused

### Environment

**MathLive version**
0.77.0

**Operating System**
macOS

**Browser**
Chrome

arnog:
Interesting... That appears to be a browser behavior. When changing tab, the document loses focus, and any focused element on the page loses focus as well.

When refocusing a page that had a focused textarea, the browsers refocus the textarea. However, they don't appear to do that for custom elements, such as `<math-field>`.

arnog:
I have implemented a workaround that mimics the browser behavior for textarea.

PR #1940, opened 2023-04-21
fix: switching tabs focus fix

mgreminger:
Previously, math-field elements would steal focus from other focused element types when switching tabs. This caused quite a bit of issues with my app since it has many math-field and text input elements on the same page and switching tabs would cause a math-field element to always gain focus, even if a text input was previously focused, making it seem like a random element is being selected. This commit fixes this issue while still making sure #1545 remains fixed (math-field elements will still keep focus when changing tabs, just like textarea elements). 

The window blur event gets triggered even when not leaving a tab (focus switching between elements on the page). By checking that `event.relatedTarget` is null before adding the focusing callback, the callback is only added when the tab is actually losing focus.

I've verified that this fix works in Firefox, Chrome, and Safari. I'm unable to create a Playwright test for this case since Playwright doesn't currently allow a tab to lose focus [Playwright #11645](https://github.com/microsoft/playwright/issues/11645).

There's still an issue when nothing is selected and the most recently selected element was a math-field, that math-field will be selected when leaving and returning to the tab. However, this fix gets the behavior closer to a textarea and solves the most significant issues I was having with focus and tab switching.

### DIFF
--- a/src/editor-mathfield/mathfield-private.ts
+++ b/src/editor-mathfield/mathfield-private.ts
@@ -470,8 +470,8 @@ If you are using Vue, this may be because you are using the runtime-only build o
     if (window === window.top) {
       window.addEventListener(
         'blur',
-        () => {
-          if (isValidMathfield(this) && this.hasFocus()) {
+        (event) => {
+          if (!event.relatedTarget && isValidMathfield(this) && this.hasFocus()) {
             window.addEventListener(
               'focus',
               (evt) => {


arnog:
Thanks for this fix. Yeah, the logic is flawed. I'll have a look.

arnog:
OK, I've made another attempt with 4deaa524c6b1be951be80c4b7af4b614672cb13e

mgreminger:
Thanks, this is a nice improvement. The [`window.removeEventHandler('blur'...)` call](https://github.com/arnog/mathlive/blob/4deaa524c6b1be951be80c4b7af4b614672cb13e/src/editor-mathfield/mathfield-private.ts#LL800C1-L800C1) is no longer needed after this change and can be removed.

arnog:
I think the change includes removing the global blur handler, unless I’m missing something.

PR #1942, opened 2023-04-21
Keyboard paste fix (fix #1937)

### LINKED ISSUE #1937: Can't paste in latex mode from firefox
`Uncaught TypeError: navigator.clipboard.readText is not a function`

Firefox does not expose `navigator.clipboard`, and tbh it's a bit annoying that Chrome asks for permission.

mgreminger:
Same logic is now used for both keyboard paste and virtual keyboard paste. Only difference is that the keyboard paste uses `event.clipboardData` to avoid needing user permission. Pasting using virtual keyboard will still require a permission dialog on safari and chrome since `navigator.clipboard` needs to be used in that case. Virtual keyboard paste will not work on Firefox since Firefox doesn't implement the `navigator.clipboard` api.

I added a test as well for keyboard copy/paste. The test can only be run on Firefox due to limitations with Playwright. However, the new keyboard paste implementation works for Chrome and Safari, as well as firefox.

I also updated the E2E tests for recent changes (`Cmd-A` select all no longer works for read-only mathfields, mouse selection is used instead, and there is a order change for subscripts and superscripts in another test).

### DIFF
--- a/src/editor-mathfield/mathfield-private.ts
+++ b/src/editor-mathfield/mathfield-private.ts
@@ -34,6 +34,7 @@ import { getDefaultRegisters } from '../core/registers';
 
 import {
   contentWillChange,
+  contentDidChange,
   deleteRange,
   getMode,
   isRange,
@@ -432,14 +433,26 @@ If you are using Vue, this may be because you are using the runtime-only build o
         onCopy: (ev) => ModeEditor.onCopy(this, ev),
         onPaste: (ev) => {
           // Ignore if in read-only mode
-          let result = this.isSelectionEditable;
+          let result = this.isSelectionEditable;          
+
+          const text = ev.clipboardData?.getData("text");
 
           if (result) {
-            result = ModeEditor.onPaste(
-              this.model.at(this.model.position).mode,
-              this,
-              ev.clipboardData
-            );
+            if (
+              text &&
+              contentWillChange(this.model, {
+                inputType: 'insertFromPaste',
+                data: text,
+              })
+            ) {
+              this.snapshot();
+              if (this.insert(text)) {
+                contentDidChange(this.model, {
+                  inputType: 'insertFromPaste',
+                });
+                requestUpdate(this);
+              }
+            } else result = false;
           }
 
           if (!result) this.model.announce('plonk');
--- a/src/editor-mathfield/mathfield-private.ts
+++ b/src/editor-mathfield/mathfield-private.ts
@@ -433,9 +433,9 @@ If you are using Vue, this may be because you are using the runtime-only build o
         onCopy: (ev) => ModeEditor.onCopy(this, ev),
         onPaste: (ev) => {
           // Ignore if in read-only mode
-          let result = this.isSelectionEditable;          
+          let result = this.isSelectionEditable;
 
-          const text = ev.clipboardData?.getData("text");
+          const text = ev.clipboardData?.getData('text');
 
           if (result) {
             if (
@@ -484,7 +484,11 @@ If you are using Vue, this may be because you are using the runtime-only build o
       window.addEventListener(
         'blur',
         (event) => {
-          if (!event.relatedTarget && isValidMathfield(this) && this.hasFocus()) {
+          if (
+            !event.relatedTarget &&
+            isValidMathfield(this) &&
+            this.hasFocus()
+          ) {
             window.addEventListener(
               'focus',
               (evt) => {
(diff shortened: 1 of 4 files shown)


arnog:
Thanks for looking into this and for this PR.
I'm not sure I understand the change in `mathfield-private.ts`. `ModelEditor.onPaste` eventually calls `contentWillChange()`, etc... so I don't get what's the difference.

It seems that they key change is that you removed the keyboard shortcut commands for Paste in `keybiding-definitions.ts`. Did you test on iPad with a keyboard after this change? As I recall, iPadOS (and I think Windows when running on a touch-enabled device?) did not send paste event when the paste shortcut was pressed, so listening for the keyboard shortcut was required to get this to work.

mgreminger:
The reason `mathfield-private.ts` needed to be updated was for the original issue seen in #1937 where pasting didn't work in Firefox when in latex mode. I just copied the logic for the [`pasteFromClipbaord`](https://github.com/arnog/mathlive/blob/ea8f3dbc62065ffd33ba8159a7452bf61366e14d/src/editor-mathfield/commands.ts#L152) command. For whatever reason, `modeEditor.onPaste` doesn't work in Latex mode, I'm not sure why, but the logic in `pasteFromClipbaord` worked, so I used that. Ideally, the `mathfield-private.ts` code should just use the `pasteFromClipboard` command directly for the DRY principle, I'll attempt to refactor the code to do that.

For the custom keyboard shortcut issue, this is to address the paste permission in Chrome, which as also mentioned in #1937. The `keydown` event happens before `paste`, so the browser doesn't recognize it as a valid paste event, thus requiring the permission dialog on Chrome based browsers (I believe Safari has a permission popup as well, though I don't see it consistently). Since Firefox doesn't have `navigator.clipboard.getText()`, the handling of the custom keyboard shortcut fails and Firefox falls back to the native paste event in `mathfield-private.ts`. I did test on the iOS simulator, and you're correct, this breaks pasting on iOS (using a keyboard, long press pasting doesn't work with or without these changes). The issue is that Safari is not recognizing a `math-field` element as a valid paste target (paste doesn't appear in the context menu and is not an option for a long press, even in the current version of MathLive). This same issue happens with Chrome as well.  A better fix will be to tell the browser that a `math-field` element is a valid paste target. I'll look into whether or not this is possible. It seems like pasting on mobile doesn't work with MathQuill either (neither long press nor keyboard pasting works). 

A paste permission requirement is a no-go for my app since this is a significant security issue so I'm pretty motivated to find a fix. It's unwise for a user to give a web page read access to their clipboard, it's a huge security risk, especially with the use of password managers that copy passwords to the clipboard. Once someone accepts the paste permission for a page, that page can read their clipboard anytime they [interact with the page](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText#security).

arnog:
OK, so the changes in `mathfield-private.ts` are probably not the right fix for #1937. `ModeEditor.onPaste()` should work for LaTeX mode, and if it doesn't that's what should be fixed.

`pasteFromClipboard` is the codepath when there is no transient use activation (i.e. the browser can't verify that the user requested the operation, for example when the user pressed the "Paste" button in the virtual keyboard). With that codepath there is limited access to the clipboard, in particular only the textual content of the clipboard, and user may need to authorize the operation.

The `onPaste` handlers are invoked in response to an event that is originating from a transient user activation (i.e. the browser can validate that the user requested the operation, and does not need to ask for permission). They have access to the full clipboard API, including all the different flavors in the clipboard. This is the preferred method to handle a Paste operation. Unfortunately, there is no way to "fake" a paste Event, or at least I didn't find one, and so the two code paths had to be separate.

I have not been able to find a way to convince Safari that a Paste operation is legal. It could be argued that it's a bug in Safari. However, the only workaround I could find was to add an interception of the keyboard shortcuts, and handling via the `pasteFromClipboard` code path, which triggers an authorization request.

I agree that the authorization request is undesirable, but I couldn't figure out a way to avoid it that works for all browsers. Note that *not* having the authorization request doesn't mean that the page does not have access to the clipboard, it simply means that the page only has access to the clipboard in response to a user action.

So, it seems like the two options are:
1/ have an authorization request in some cases
2/ have no authorization request, but pasting won't work at all on iPad

If for your application you feel that (2) is preferable, I would suggest that a way to deal with that would be to customize the keybindings and remove the ones for `ctrl/cmd`+`V`.

Maybe you can find some way to convince Safari that the paste operation is applicable when a mathfield is focused. If so, great, and the clipboard code path that doesn't require authorization can be avoided and the keybindings removed.

mgreminger:
Thanks for the clarification on `ModeEditor.onPaste()`, I wasn't aware of the difference when handling a native paste event in how it used the clipboard object data beyond the text data. So, yes, you're correct, this doesn't fix #1937 correctly. 

I'll close this pull request and create a new one with the Playwright test fixes since the test fixes were unrelated to this pull request anyway.

PR #1952, opened 2023-04-27
perf: defer getBoundingClientRect calls

mgreminger:
#1926 introduced a change where `getBoundingClientRect` was being called in `makeProxy`. `makeProxy` is triggered frequently during math-field and DOM updates (and during initial math-field creation). `getBoundingClientRect` is an expensive function and the frequent calls were causing performance issues during DOM updates and during page load for pages with a large number of math-field elements. Since `getBoundingClientRect` is only needed to place the environment popover when toggling the virtual keyboard, this change defers the call to `getBoundingClientRect` until it is  needed to place the popover. The existing functionality is unchanged.

### DIFF
--- a/src/public/virtual-keyboard.ts
+++ b/src/public/virtual-keyboard.ts
@@ -225,7 +225,7 @@ export interface MathfieldProxy {
   readonly mode: ParseMode;
   readonly style: Style;
   readonly array?: unknown;
-  readonly boundingRect?: DOMRect;
+  readonly field?: HTMLElement;
 }
 
 /**
--- a/src/virtual-keyboard/mathfield-proxy.ts
+++ b/src/virtual-keyboard/mathfield-proxy.ts
@@ -9,7 +9,7 @@ export function makeProxy(mf: MathfieldPrivate): MathfieldProxy {
     canRedo: mf.canRedo(),
     style: mf.selectionStyle,
     array: mf.model.parentEnvironment,
-    boundingRect: mf.field?.getBoundingClientRect(),
+    field: mf.field,
     mode: mf.mode,
   };
 }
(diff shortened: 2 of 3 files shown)


mgreminger:
Hold off on merging this one, doesn't quite fix the issue. Turns out `updateEnvironmemtPopover` gets called just as often as `makeProxy` so this doesn't change anything. I'll look into why `updateEnivornmentPopover` is getting called so frequently.

arnog:
Unfortunately, this doesn't work: the `MathfieldProxy` can only contain properties that can be `structureClone`'d. Since the keyboard and the mathfield could be in different frames, the `HTMLElement` may not be accessible from the document in which the keyboard is.

The good news is that the environment popover really has nothing to do with the virtual keyboard, so it should not depend on the `MathfieldProxy`. Let me know if you're interested in looking at fixing that, otherwise I'll have a look at it.

mgreminger:
I'm not sure how to fix this since the implementation of #1926 seems to tie the rendering of the environment popover to the virtual keyboard. Given how it's implemented, I don't think it would work in an iframe anyway since the boundingRect is calculated in the math-field context and would be in a different window coordinate than the environment popover since it seems to be placed in the virtual keyboards DOM structure.

I don't use the environment popover but #1926 has killed the performance of my app, which can have 100's of math-field instances on the same page. In particular the call to [`getBoundingClientRect`](https://github.com/WildYellowfin/mathlive/blob/eea12ae71702faa2b1763f543891716f74be101a/src/virtual-keyboard/mathfield-proxy.ts#LL12C8-L12C8) is very expensive and gets called all the time, even when not using the environment popover or the virtual keyboard.

arnog:
@mgreminger ok, I've checked in a fix. I've rearchitected the environment panel to not be tied to the virtual keyboard, so it is no longer necessary to include the boundsrect in the proxy.

Performance was an issue while resizing, right? Or were there other use cases where it was problematic? I have tried a test case with thousands of mathfields in a page and performance seems fine to me now, but let me know.

mgreminger:
@arnog, that solved the issue, many thanks! I was mostly seeing the issue when there were DOM changes, like when dragging to reorder math-fields, which would trigger the makeProxy on many of the elements, making the drag operation sluggish. These updates have solved the issue. Thanks!

PR #1953, opened 2023-04-27
Locale dependant mathfield

zeyad-ahmad-aql:
I faced a case where I will be using multiple mathfields in the same page, but wanted their locales to be different, intead of 1 locale to rule them all.

Please review and tell me if you have any comments. Also feel free to reject this merge request if you feel that it won't be benifical for you.

What I did was the following:
- Rename `locale` in the `l10n` object to `gLocale` (global locale)
- Add `local` to `MathfieldOptions`
- Add logic to get current focused mathfield from the `MathfieldElement` class

If a Mathfield doesn't have a locale, the `l10n` object's `gLocale` is used.

On switching between mathfields with different locales, First, I thought about updating the layout toolbar locale only instead of rebuilding the whole virtual keyboard. But I ended up with rebuilding the keyboard just to be safer, maybe some keycaps will have their own locale in the future for example.

When I tried to call the `VirtualKeyboard.singleton.rebuild` method to rebuild the keyboard on mathfield locale changes, the behavior wasn't quite right. So, I ended up calling the `hide`, `show`, and `update` methods Instead.

---

Other secondary changes: 
- Fix virtual keyboard toolbar tooltips not showing due to overflow hidden
- Add test page for multi locale mathfields
- Add missing redo tooltip
- Add 2 missing tooltips in `ar` locale
- Add `ML__layouts-toolbar` class to layouts toolbar markup. I just felt like it can be needed later
- Add `isConnectedToVirtualKeyboard` to `MathfieldElement` class. I just felt like it can be needed later

### DIFF
--- a/css/virtual-keyboard.less
+++ b/css/virtual-keyboard.less
@@ -1,4 +1,3 @@
-
 //
 // Breakpoints
 //
@@ -17,7 +16,7 @@
 @shadow4: 0 14px 28px rgba(0, 0, 0, 0.25), 0 10px 10px rgba(0, 0, 0, 0.22);
 @shadow5: 0 19px 38px rgba(0, 0, 0, 0.3), 0 15px 12px rgba(0, 0, 0, 0.22);
 
-// 
+//
 // blue-50   #e2f0fd
 // blue-100   #b6d9fb
 // blue-200   #86c0f9
@@ -86,7 +85,7 @@
 
   --keycap-shift-font-size: 16px;
   --keycap-shift-color: var(--keyboard-accent-color);
-  --environment-shadow: rgba(0,0,0,0.4);
+  --environment-shadow: rgba(0, 0, 0, 0.4);
 }
 
 .if-can-undo,
@@ -212,7 +211,7 @@ body > .ML__keyboard.is-visible.animate > .MLK__backdrop {
 
   background: var(--keyboard-background);
 
-  overflow: hidden;
+  // overflow: hidden;
 }
 
 .backdrop-is-transparent .MLK__backdrop {
@@ -412,7 +411,7 @@ body > .ML__keyboard.is-visible.animate > .MLK__backdrop {
       line-height: 0.8;
       font-size: calc(min(1rem, var(--keycap-small-font-size, 16px)));
       font-weight: 600;
-      padding: 8px 12px 8px 12px;      
+      padding: 8px 12px 8px 12px;
       &:hover {
         background: var(--keycap-secondary-background-hover);
       }
@@ -423,7 +422,7 @@ body > .ML__keyboard.is-visible.animate > .MLK__backdrop {
       color: #ddd;
     }
     .action.primary:hover {
-      background: #0d80f2 ; // blue-500
+      background: #0d80f2; // blue-500
       color: #fff;
     }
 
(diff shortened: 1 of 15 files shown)


arnog:
Until recently, the locale-related properties used to be properties of the mathfield instances, but I moved them to be global properties of the `MathfieldElement`. I wasn't sure at the time if there was any interest in having mathfields with different locales in the same page. I guess you answered that question.

That said, I also tried to consider precedent. For example, the locale of the `<input type="date">` widget is determined based on the page or system locale. 

On the other hand, the W3C recommends:
>Implementations SHOULD present non-linguistic fields in a document format or application using a format consistent with the language of the surrounding content and are encouraged to provide controls which are [localized](https://www.w3.org/TR/ltli/#dfn-localization) to the same [locale](https://www.w3.org/TR/ltli/#locale) for input or editing.

(https://www.w3.org/TR/ltli/#ltli-impl-input-like-doc-language)

So, adding support for per-mathfield localization would be consistent with that. For consistency with other HTML elements, I would use a `lang` attribute, rather than a `locale` attribute.

There are a number of properties that are influenced by the locale, including the user readable strings, but also the layout of the alphabetic keyboard (AZERTY vs. QWERTY), the decimal separator (`,` vs `.`) and the navigation order of numerator and denominator (for some east-asian languages).

Your solution currently handles some UI strings, but I believe the rest should be handled as well.

A few other comments:
- I don't understand the purpose of renaming `l10n.locale` to `l10n.gLocale`. Unless I'm mistaken, there can't be a namespace collision, so that doesn't seem like a necessary change.
- Having the argument to `localize()` be a MathfieldElement seems unnecessary, since you always pass the locale string when invoking this function. In general, I would avoid creating dependencies between different layers of the architecture, especially circular dependencies.
- Similarly, the `MathfieldElement.current` property seems superfluous. It seems like it would be redundant with `document.activeElement`.
- The `localize()` function has some logic to first find a match to the entire locale, i.e. "fr-CA", then to fall back to a language only, i.e. "fr", then to english. This logic is bypassed when the locale is passed as an argument to `localize()` but I don't think it should.
- I would avoid adding `ML__layouts-toolbar`. In general, I try to avoid adding CSS classes unless absolutely necessary, as they are a burden to maintain.
- I also would avoid `isConnectedToVirtualKeyboard` unless there's a clear use case for it.

Let me know if you're interested in making some changes to account for this feedback. I understand if you're not interested, in which case I would have a look at it. Either way, thanks for raising this issue, and for your contribution towards solving it.

arnog:
Another alternative to consider:

```js
document.querySelector('math-field').addEventListener('focusin', (evt) => {
   const lang = evt.target.dataset.lang;
   if (lang) MathfieldElement.locale = lang
});
```

This would dynamically change the locale when a mathfield is focused. The `MathfieldElement.locale` property would react by changing the `l10n.locale` and updating the UI as necessary (rebuilding the keyboard, for example). Alternatively, that property could be `mathVirtualKeyboard.locale`.

zeyad-ahmad-aql:
Thank you for your feedback. In short, I found a case where I needed a specific feature and I wanted to get to know the library a little better, so I decided I would develop it and see what's the best approach I can come up with. So this is a suggestion. Please feel free to take whatever You think is compatible with your ideas and architecture, leave whatever you think it isn't.

I would suggest you cherry pick the follwing commits though:
- https://github.com/arnog/mathlive/pull/1953/commits/a9700350930eb43131f0d2e65e298ea32353c915
- https://github.com/arnog/mathlive/pull/1953/commits/74f101fc65963b5274c60274e6e78bd1982c12df
- https://github.com/arnog/mathlive/pull/1953/commits/acc54b2918e9817a1047e2bdc37354fb493489da

Tbh, I don't think I will try to invisigate further in this. I'm sorry I can't put some more effort into this. But the `lang` attribute idea can be helpful.

Again, thank you for your feedback.

Issue #1357, opened 2022-02-23
How to "shift" between custom virtual keyboard layers


wlewis-formative:
A colleague and I are attempting to add a custom virtual keyboard similar to the default "roman" keyboard.
We'd like the `Shift` key to behave identically: that is, the "lower" layer should be visible by default, and when
a user clicks `Shift`, the keyboard should switch to the "upper" layer (and remain there until they click `Shift`
again). We've tried adding various `class`es to the `Shift` key, but to no avail. We've also considered the
possibility that a custom "shift handler" is required, but don't know where to install it.

Is it possible to produce this effect in a custom keyboard at this time, and if so, could you provide a brief example
or pointers on how to do so?

PS Thanks for this great package.

arnog:
Thank you, glad you enjoy MathLive.

It's possible, but it's a bit involved and I have written only limited documentation about this. Here are a few tips...

To have a key that triggers a "shift", use something like this:
```json
{
   "label": "&#x21e7;", 
   "class": "layer-switch shift modifier font-glyph bottom left w15'"
},
```

The `&#x21e7` is the Unicode glyph for the shift-key, but you can use whatever is appropriate for your use case. 

The set of classes are mostly decorative, except for the `layer-switch` class and the `shift` class. The `layer-switch` class is the the one that indicates that this keycap triggers a layer switch.

Now, there are two "modes" of keyboard shifting. 

The first is a temporary state, which you get if you press and hold the shift key. When you release the shift key after this press and hold, you revert to the original state. In this temporary state the label (and commands) associated with the keycaps on the current layer are modified. The label of the keycaps with a `shifted` property are changed to the content of the `shifted` property. If they have a `shiftedCommand` property, that is the command that will be executed if the keycap is pressed in this state. If the label of the keycap is a single lowercase letter (`a-z`), it is automatically turned to its uppercase counterpart,  even if there is no `shifted` property.

This first state is triggered by the `shift` class. If you don't want this behavior, but simply want to switch to another layer, include the `shift-layer` class, but not the `shift` class.


 Incidentally, while testing this I discovered an issue: the keycap doesn't properly revert to its original state if it has a `shifted` property and content other than a single "a-z" character. I'll push a fix for this shortly.


[...]

PR #1954, opened 2023-04-30
fix: switchKeyboardLayer

eagleoflqj:
#1357 
This seems a regression after 0.90.
Test case provides the new way to switch keyboard layer by shift.

### DIFF
--- a/src/virtual-keyboard/commands.ts
+++ b/src/virtual-keyboard/commands.ts
@@ -3,32 +3,14 @@ import { register } from '../editor/commands';
 import { hideVariantsPanel } from './variants';
 import { VirtualKeyboard } from './virtual-keyboard';
 
-export function switchKeyboardLayer(layerName: string | null): boolean {
+export function switchKeyboardLayer(layerName: string): boolean {
   const keyboard = VirtualKeyboard.singleton;
   keyboard.show();
 
   // If the variants panel was visible, hide it
   hideVariantsPanel();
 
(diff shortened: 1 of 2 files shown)


arnog:
Thanks for the PR. Could you tell me more about what was broken? How should I test this? BTW, as a side note, using the `shift` property is the recommended way to deal with multiple values for a key. Only use layers if you need a different layout (physical arrangement of the keycaps) between layers.

eagleoflqj:
The problem with `[shift]` is that I can't release the key when I want it to take effect. It's not possible to use it on mobile phone with single hand.
So as you can see in the e2e test I added, I define a layout with 2 layers - one for lower letters and the other for upper letters. Press shift key to switch and no need to hold it.
The bug with `switchKeyboardLayer` is that it expects `layer.dataset.layer` to be the name of the layer. However with current mathlive, the element doesn't contain `data-layer` attribute as it did in 0.89. So https://github.com/arnog/mathlive/blob/2ba472eed955bc8227d45381d0c2e579190aaec9/src/virtual-keyboard/commands.ts#L17 doesn't work.
About how to test it, in any web page with a `MathFieldElement`, assign `window.mathVirtualKeyboard.layouts` as in the test case. The shift doesn't work in master's build, but works in this branch's build.

arnog:
Oh, I see. OK, thanks for the explanation and for tracking this down. That's a good fix.

That said, it is a bug that the shift key when used with the `shift` property only works while pressed. It should also work as a toggle. I'll look into it.

Issue #28, opened 2017-10-22
Need a way to set a minimum font size


NSoiffer:
## Description
As you use exponents, fractions, etc., the rules of TeX shrink the font size. In TeX, it will shrink at most three times.  mathlive seems to follow this rule, but the smallest size is too small to easily read on a screen.

Below is an example (see exponent in denominator):
![image](https://user-images.githubusercontent.com/1545836/31858642-93acf3ac-b6b1-11e7-9752-030d7f3f6bba.png)

For people with good sight, this is very hard to read. Any one with the slightest visual disability would not be able to read this. If another app is using mathlive, then for accessibility reasons, the app may want to sacrifice some typesetting quality for readability. 

Note: MathML has an attribute to deal with the readability issue. From the [MathML ref](https://www.w3.org/TR/MathML/chapter3.html#presm.mstyle.attrs):

> scriptminsize | length | 8pt
> 
> Specifies the minimum font size allowed due to changes in scriptlevel. Note that this does not limit the font size due to changes to mathsize. See Section 3.1.6 Displaystyle and Scriptlevel.

It would be helpful to have such a setting in mathlive.

arnog:
This is an interesting problem. It turns out this is one of those things that are near impossible to implement with CSS. Sigh.

However, while investigating this, I discovered two problems:
- the font size of the enclosing block was ignored and the size hard-coded to 16px. We've talked about workaround for this before (using `scale()`), but I've actually found the problem and fixed it. As a result of this fix, you may need to explicitly set the font-size attribute of the block enclosing a MathLive formula to at least 16px.

- The sizing for the case above (fraction inside a supsub) was calculated incorrectly and the size reduced twice. This is now fixed. As a result, by default, the formula will be more legible.

This doesn't address directly the issue of having some control over the minimum size. This is a difficult problem because the sizing is relative, and there's no simple way to put a cut off for a minimal size programmatically. However, the good news is that browsers do have a setting specifying the minimum font-size (in Firefox: Preferences > Font & Colors > Advanced > Minimum Font Size, in Chrome: Settings > Customize Fonts > Minimum Font Size). This settings will now be respected.

I hope this address the issue. Re-open the report if it does not.

PR #1955, opened 2023-05-02
feat: add minFontScale property

mgreminger:
Relates to #28. By default, MathLive reduces the font size down to half of the `math-field` element font size for doubly nested scripts or fractions. This is a good default, and matches the behavior of other Latex renderers such as MathJax. However, there are instances where cutting the font size in half makes equations illegible if the starting font element font size is below a certain threshold. For example, a element font size of 16px results in the following: 

before

In some applications it is not practical to use a large font size for the math-field in order to ensure that the half font size is still legible. In these cases it would be desirable to limit how much the font can be scaled down. This pull request implements a `minFontScale` property (also accessible as the `min-font-scale` attribute) that limits how much the font can be scaled down. The default is `0`, which allows the `math-field` to use its default logic. The following example uses a 16px `math-field` font size with `minFontScale=0.75`:

after

I followed the logic of existing properties/attributes in implementing this new property. Let me know if there are any questions or if any additional updates are needed.


arnog:
Thanks for the PR. Looks good, and it's a reasonable approach to this issue.

PR #1961, opened 2023-05-11
Mimic mobile keyboard shift lock

zeyad-ahmad-aql:
I wanted to mimic the mobile keyboard's shift lock behavior.
- When you press shift once, it is locked, the next inserted character only will be inserted in uppercase, and the shift lock will reset.
- if pressed twice, it is locked, all the characters will be inserted in uppercase, and the shift lock will remain.
- if pressed trice, the shift lock will reset.

Other commits:
1. There was an issue: On switch virtual keyboard layer while the shift is pressed, the new layer keycap's aren't shifted. My fix was to call the `keyboard.render` method on setting the virtual keyboard's `currentLayer` attribute if the keyboard is shifted.
2. I also did something if you don't feel it's the behavior you need, feel free to revert it. if virtual keyboard shift is locked, insert the characters that were entered by the physical keyboard in uppercase.

### DIFF
--- a/src/editor-mathfield/keyboard-input.ts
+++ b/src/editor-mathfield/keyboard-input.ts
@@ -27,6 +27,7 @@ import { showKeystroke } from './keystroke-caption';
 import { ModeEditor } from './mode-editor';
 import { insertSmartFence } from './mode-editor-math';
 import type { ParseMode } from '../mathlive';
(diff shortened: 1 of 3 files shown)


zeyad-ahmad-aql:
Also add fix for caps-lock-indicator not displayed if enabled on showing virtual keyboard

arnog:
Excellent! Thank you.

PR #1964, opened 2023-05-12
fix: prevent errors when embedded in iframe

mgreminger:
After switching EngineeringPaper.xyz over to mathlive from mathquill, I discovered that my blog posts with EngineeringPaper.xyz iframe embeds were no longer functional since MathLive fields are not functional when embedded in an iframe on an outer page that does not itself contain a math field.

See an example [page here with mathlive fields in an iframe](https://15d141e0.ep-blog.pages.dev/lead-screw-torque-and-efficiency-calculations) (scroll down to see the iframe).

[Here's the same EngineeringPaper.xyz sheet](https://engineeringpaper.xyz/dRttcDo7GXeCD2iHLc5R7x) rendered outside of an iframe using MathLive.

This is the [blog page rendered correctly](https://blog.engineeringpaper.xyz/lead-screw-torque-and-efficiency-calculations) (falling back to the MathQuill implementation).

The issue was the  `_boundingRect` static member of the `VritualKeyboardProxy` class being used before being initialized. This change initializes the `_boundingRect` value, which fixes the iframe issue.

I also added a Playwright test to test the virtual keyboard targeting a mathfield in an iframe to ensure that this change does not introduce a regression with using the virtual keyboard to target a math field in an iframe.


arnog:
Oops. Sorry about that. Good catch, thank you!

Issue #1815, opened 2023-02-08
Visualization of empty groups


ichibha:
This issue is derived from #1812 . Is it possible to display something for empty groups, which is not currently implemented?

scytacki:
Hi @arnog I'm working with @dennisrcao on this. Is this https://github.com/concord-consortium/mathlive/pull/2 a good place to add a visualization of empty groups?

arnog:
Great! Thanks for looking into this. Here’s how I would approach this:
- modify `Box.toMarkup()` to add a class (`ML_empty-group`?) if the group is empty and does not contain the caret (if the caret is in an empty group I would just do nothing since at that point the caret makes it clear where you are. The point of making empty groups visible is to draw attention to them when they could otherwise be invisible and difficult to click on)
- Have that class do some styling like setting a minimum width/height, and perhaps a background color. The class should only be active if the mathfield has the focus
- Trigger that behavior off a new configuration option

scytacki:
To clarify, are you saying we should not identify the empty group in `createBox` and instead identify it in toMarkup? 

It seemed to me that identifying in `createBox` was better since it is closer to the "model".
Identifying it in `toMarkup` meant looking for a box that had no siblings, no children, and had an id. The last two parts are easier in `createBox` because we can just look for a `first` atom. After identifying this atom, the createBox function could return a box with a property indicating it is a empty group (or a new subclass of Box could be used). So then toMarkup could add the class to it.

arnog:
You’re right, you still need to identify it as an empty group, and `createBox` is the best place for it. I would either create a Box subclass, or you might just be able to create a Box with a custom class.

PR #1977, opened 2023-05-22
185091247 - Placeholders are shown for null/empty operands

dennisrcao:
WIP
Do not review


arnog:
FYI, you might want to consider [1815](https://github.com/arnog/mathlive/issues/1815). The idea is to have an option so that empty groups would be made more prominent (given some width, and a background color). An empty group being different from a placeholder, as you can have valid content with empty groups (i.e. "2^{}") which doesn't have a placeholder.

PR #1987, opened 2023-05-24
fix: selection highlight only for focused math-field

mgreminger:
As mentioned in the gitter chat, this addresses selection highlighting showing for unfocused math-field elements after changes for 0.94.5


arnog:
Yep, that works. Thanks!

PR #2022, opened 2023-06-17
fix: pointer selection after using parentArray

### LINKED ISSUE #2016: Can no longer use pointer to highlight atoms after using "parentArray" function
## Description

When an environment is created from the `parentArray` function (`src/editor-model/array.ts`) e.g. when using the hotkey <kbd>alt/option</kbd>+<kbd>Return</kbd> it only allows you to select the entire parent atom with the mouse pointer, and also copy ( <kbd>ctrl/cmd</kbd>+<kbd>c</kbd> ) no longer works.

## Steps to Reproduce

1. visit https://cortexjs.io/mathlive/
2. Type some things in
3. Press alt + enter to create a new line / array environment
4. try to highlight the contents with your mouse, or try to copy the contents to clipboard

### Actual Behavior

No individual elements can be highlighted (only the entire environment), and copying does not work.

Note: Highlighting elements still work when you paste in the environment instead of using the hotkeys
```
\begin{split}a+b=c\\ 1+2=3\end{split}
```

### Expected Behavior

I should be able to highlight elements in the environment, as if it were like a regular matrix.

### Environment

**MathLive version**

0.94.7

**Operating System**

Windows 10.0.19044 Build 19044

**Browser**

Occurs in Chrome / Firefox / Edge Chromium

manstie:
closes #2016

### DIFF
--- a/src/editor-model/array.ts
+++ b/src/editor-model/array.ts
@@ -164,13 +164,13 @@ function parentArray(
       let array: ArrayAtom;
       if (where.endsWith('column')) {
         array = makeEnvironment('split', [[firstCell, secondCell]]);
-        model.root = array;
+        model.root.body = [array];
         if (isPlaceholderCell(array, 0, 0)) selectCell(model, array, 0, 0);
         else if (isPlaceholderCell(array, 0, 1)) selectCell(model, array, 0, 1);
         else model.position = model.offsetOf(cursor);
       } else {
         array = makeEnvironment('lines', [[firstCell], [secondCell]]);
-        model.root = array;
+        model.root.body = [array];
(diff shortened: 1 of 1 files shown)


manstie:
I realised this allows the cursor to go "outside" the environment, so if that is undesired I am unsure how to fix this, since there are quite a few places which rely on `type === 'root'`.

arnog:
I need to investigate this. I'm not sure what's causing the problem, but I think there needs a different solution.

arnog:
Fixed with https://github.com/arnog/mathlive/commit/e9925484eeb5715db061d96475e3fef934496c3a

PR #2025, opened 2023-06-19
fix: added closing angle bracket in LaTeX to MathML conversion

ainhoa-basadre:
In 'convertLatexToMathMl(...)' added closing angle bracket to '\<mo>' when left delim added


arnog:
Ooops.. Good catch. Thanks for the PR.

PR #2033, opened 2023-06-25
Virtual Keyboard display on older browsers

### LINKED ISSUE #2028: [Bug] The virtual keyboard layout is broken on iOS
## Description

https://github.com/arnog/mathlive/assets/44936298/a1a2b998-92ad-46aa-96f9-0b0a789b29bd

### Environment

**MathLive version**: Newest 

**Operating System**:  **iOS 15.7.3 (iPhone 6s)** 

**Browser**:  Safari

manstie:
The keyboard does not display properly on browsers which do not support `cqw` as it is a relatively new feature. A friend had Chromium 103 installed when the minimum is 105 for this feature.

The good news is in the majority of places it is used for setting the width to a percentage width of the containing element, which `%` replaces seamlessly.

There still remains a 4cqw inside of `--_keycap-font-size` which can not be simply replaced with % because that scales relative to font size. On browsers which do not support `cqw` it will default to inherit.

closes  #2028

### DIFF
--- a/css/virtual-keyboard.less
+++ b/css/virtual-keyboard.less
@@ -379,22 +379,22 @@ body > .ML__keyboard.is-visible.animate > .MLK__backdrop {
     }
     .w5 {
       width: calc(
-        ~'min(var(--_keycap-max-width, 100px), 10cqw)' / 2 - var(--_keycap-gap);
+        ~'min(var(--_keycap-max-width, 100px), 10%)' / 2 - var(--_keycap-gap);
       );
     }
     .w15 {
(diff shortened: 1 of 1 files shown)


arnog:
Unfortunately I had to revert this change. `10%` and `10cqw` are not equivalent. The percentage is a percentage of the parent element, not of the container. In many cases, especially on mobile, the layout would end up incorrect.

If support for older browser is still desired, a possible workaround for this would be to calculate the correct width via JavaScript and set a CSS variable to capture it.

PR #2036, opened 2023-06-29
fix: typo in contains-highlight-background-color css variable

mgreminger:
Fixes typo in css variable name


arnog:
Good catch. Thank you!

PR #2043, opened 2023-07-02
fix: f, g, and h no longer impact behavior of /

### LINKED ISSUE #1869: Different behaviour of / depending on the letter that precedes it
## Description

When typing / to create a fraction in mathlive the result differs depending on what comes before. All numbers will result in `number/placeholder` while some letter will be `letter/placeholder` while others will be `letter * placeholder/placeholder`. 

For example typing `r` then `/` will result in `r/placeholder` while typing `f` then `/` will result in `f * placeholder/placeholder`. 

So for example if you're to write an expression for the area of a triangle of base=b and height=h then you get different behavior depending on the order you write it in. `A=hb/2` will result in `A=(hb)/2`, while `A=bh/2` will result in `A=bh*2/placeholder`.

_Arno wrote on Gitter_
> "f", "g" and "h" are considered functions which affects how they are handled. However, one could argue that the current behavior is incorrect. When trying to find an implicit argument for the numerator, functions are excluded, which is why it looks left and stops at "f", "g" or "h". This behavior is intended to deal with the case \cos 2 followed by /. In most cases you want the 2 in the numerator, but not the \cos. However, that works because \cos is a special function that can be written without parentheses around its arguments. That should not apply for "f", "g" and "h".

Note: This was just a curiosity that I found, not sure it's the most important of issues. :)

mgreminger:
This pull request removes special treatment of f, g, and h as implicit functions while typing in expressions. The main side affect of the old behavior was that the divide command (typing /) worked differently for variables containing f, g, or h as compared to those without f, g, or h. Fixes #1869

In EngineeringPaper.xyz, the user needs to type in unit names, many of which contain f, g, or h. When typing a unit name with one of these letters, the divide command behavior is surprising, see the screen recording below:

https://github.com/arnog/mathlive/assets/6439649/c4cc79c9-bcdf-475b-813c-cddd9f253667

After this change, the behavior is consistent as with variables names that don't contain f, g, or h:

https://github.com/arnog/mathlive/assets/6439649/82ed75c0-bd15-474d-8563-88bd9322ab85

In this pull request, I simply removed the special treatment of f, g, and h. All of the current tests still pass. There doesn't seem to be other side affects besides the change in the behavior of the divide command. For example, the math json output doesn't seem to be impacted.

There's the possibility, of course, that I'm missing a side affect of this change. In that case, it may be desirable to keep the current behavior as the default and to provide the user the option to change which characters are treated as implicit functions.  If you would like me to implement this as an option, let me know, and I would be happy to implement. I was thinking it would be best to add to the `InlineShortcutsOptions` object as maybe `implicitFunctions`.

### DIFF
--- a/src/core-definitions/definitions-utils.ts
+++ b/src/core-definitions/definitions-utils.ts
@@ -922,17 +922,6 @@ export function getDefinition(
     };
   }
 
-  // Special case `f`, `g` and `h` are recognized as functions.
-  if (
-    info &&
-    info.definitionType === 'symbol' &&
-    info.type === 'mord' &&
-    (info.codepoint === 0x66 ||
-      info.codepoint === 0x67 ||
-      info.codepoint === 0x68)
-  )
-    info.isFunction = true;
-
   return info ?? null;
 }
 


mgreminger:
I'm closing this pull request since the underlying issue has been resolved in 0.98.5. Thanks for the fix.

Issue #2053, opened 2023-07-10
mathfield captures touch pointer events when unfocused


IBastawisi:
## Description

the mathfield should only handle pointer events after it receives focus when touch input is used to allow the user to scroll the page.

### Actual Behavior

The field captures the `pointerdown` event and prevents scrolling the page.

https://github.com/arnog/mathlive/assets/13211445/f526c54b-2659-4a98-8890-2f15e8c39579

### Expected Behavior

When unfocused, the expected behavior is scrolling the page.

https://github.com/arnog/mathlive/assets/13211445/81a74039-da6d-4bc7-84d5-eba1fd7236b9

### Environment

**MathLive version**: **0.95.0**

PR #2054, opened 2023-07-10
disable touch pointer events if host isn't focused

IBastawisi:
This PR fixes https://github.com/arnog/mathlive/issues/2053


arnog:
Thanks for the PR.

Issue #2065, opened 2023-07-20
Pressing `Shift` on physical keyboard ideally should not have side effects or "toggle" virtual keyboard


oscarhermoso:
## Description

This is some feedback on what I believe was a feature from [0.94.0](https://github.com/arnog/mathlive/releases/tag/v0.94.0)

> Improved behavior of the Shift key in the math keyboard. Single-press
the Shift key to set it temporarily, double-press it key to lock it (similar
to CapsLock), triple-press it to unlock. This is similar behavior to the
ones of mobile virtual keyboards.

While this works well for mobile virtual keyboards, where users are "tap typing" and looking at the virtual keyboard, users with physical keyboards find this behaviour buggy/unexpected.

Here are some unexpected behaviours with the current version ([0.94.7](https://github.com/arnog/mathlive/releases/tag/v0.94.7)) of MathLive when using a physical keyboard on https://cortexjs.io/mathlive/:

### Steps to reproduce
- User presses `Shift` once and releases the key
- User types `abc` quickly

### Expected
If the user is familiar with MathLive `Shift` toggling, they will expect `Abc`

If they are a PC user and using MathLive for the first time, they probably expect `abc`, because they don't expect `Shift` to toggle.

### Actual
- Math field may contain `ABC`, `ABc` or `Abc` depending on how fast they were typing

### Steps to reproduce
- User intends to type `a * b = c` on their physical keyboard
- Type `a`
- Hold `Shift`, but then release immediately (perhaps user was distracted for a second, or changed their mind about including a bracket in their equation)
- Type `Shift` + `8` to enter an `*`
- Press `b`

### Expected
Math field contains `a * b`

### Actual
Math field contains `a * B`, and virtual keyboard is "locked in"

### Steps to reproduce
- User is on a physical keyboard, and is familiar with current MathLive shift to toggle while the virtual keyboard is open
- They collapse the virtual keyboard
- They press `Shift` and release
- They press `=`

### Expected
- They expect to see `+` (`Shift` + `=`), same as 

### Actual
- They see `=`

### Steps to reproduce
- User types `Shift` + `a`, but continues holding shift
- They then type `Shift` + `b` and continue holding shift

### Expected
- Mathlive keyboard continues to show `Shift` being held, Mathfield contains `AB`

### Actual
- Mathlive keyboard no longer showing `Shift` is held, Mathfield contains `AB`

## Ideal outcome (in my opinion)

[...]

Issue #2064, opened 2023-07-20
Holding `Shift` on Windows physical keyboard rapidly toggles `Shift` on virtual MathLive keyboard


oscarhermoso:
## Description

On Windows, holding `Shift` with a physical keyboard rapidly toggles the `Shift` modifier on `MathLive` keyboard

## Steps to Reproduce

On MS Windows 

1. Visit MathLive Demo at https://cortexjs.io/mathlive/demo/
2. Hold `Shift`

### Actual Behavior

Keyboard rapidly toggles, see gif below

![mathlive-shift](https://github.com/arnog/mathlive/assets/23239955/c90b036e-43e3-4d9f-bf83-65420afd1992)

### Expected Behavior

Holding `Shift` should enable "once" when held, until it is released, and then ideally return to an unmodified keyboard.

(although I also understand it may be intended functionality to remain "toggled on" once released)

### Environment

MathLive version on website 2023-20-07, assume version 0.94.7

Windows 10 Home, v10.0.19045 Build 19045

Both Chrome 115.0.5790.99 (64-bit) and Firefox Version 115.0.2 (64-bit)

PR #2069, opened 2023-07-21
feat: Press/hold `Shift` modifier instead of toggle when using with physical keyboard, and fix rapid toggling of `Shift` on Windows

oscarhermoso:
Fixes https://github.com/arnog/mathlive/issues/2064
Fixes https://github.com/arnog/mathlive/issues/2065

FYI @zeyad-ahmad-aql, this changes part of your previous pull request https://github.com/arnog/mathlive/pull/1961

Would you like a config option so that physical keyboards still have a toggling `Shift`?


arnog:
Thanks for the PR. Looks good!

PR #2103, opened 2023-08-22
fix: virtual-keyboard keycap hover color variable

AllanChain:
I think it's a typo because the first occurrence is correct:

https://github.com/arnog/mathlive/blob/6175225fe2a26b7f1a4da5b1daec245045bf293d/css/virtual-keyboard.less#L62-L63

This simple fix will enable configuring the hover background color of keycaps.


arnog:
Yes, thank you!

PR #2137, opened 2023-09-15
fix: remove duplicate virtual keyboard selector in dark mode

idealsh:
Fixed a typo that causes the keyboard to adapt to system theme even when `theme="dark"` is set.


arnog:
Good catch! Thanks for the PR!

PR #2187, opened 2023-11-29
feat: accept-all option for autocomplete

AllanChain:
By default, when inputting LaTeX commands, accepting suggestion is done by `[Tab]` (`accept-suggestion`), and exiting to math mode is done by `[Enter]` (`accept`).

This PR adds a `accpet-all` option to autocomplete, which does both `accept-suggestion` and `accept`. If there are no suggestions, only `accept` is done. This allows users to bind keys to accept suggestion and exit latex mode at the same time, which is convenient. (which is also consistent with LyX if you know it)

### DIFF
--- a/src/editor-mathfield/autocomplete.ts
+++ b/src/editor-mathfield/autocomplete.ts
@@ -121,24 +121,29 @@ export function acceptCommandSuggestion(model: ModelPrivate): boolean {
  */
 export function complete(
   mathfield: _Mathfield,
-  completion: 'reject' | 'accept' | 'accept-suggestion' = 'accept',
+  completion:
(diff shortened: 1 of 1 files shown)


arnog:
Thanks for the PR. That's an interesting idea. 

Do you think a default keybinding should use that option?

AllanChain:
While I'm conservative for behavior changes, I personally appreciate the "`Tab` to cycle and `Enter` to accept and exit to math" way.

And I have to point out that if it is made default, there are some rare edge cases where the user intended `\al` (just an example) and the suggestions only list `\aleph` etc., and the user is impossible to input `\al`.

PR #2197, opened 2023-12-08
Add npm installation instructions to README

hsellik:
People are confused about installing and using this library with npm and plain javascript. I was especially confused about the initialization part.

Adding this information to the main README helps out people who visit the npm registry and want to get started quickly without diving deep into the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,22 +49,35 @@ And the best part? It's easy to customize to your needs!
 
 ## 🚀 Getting Started
 
-Using MathLive is easy! Simply add a `<math-field>` tag to your page, and it
-works just like a `<textarea>` or `<button>` element. You can manipulate the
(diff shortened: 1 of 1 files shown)


arnog:
Thanks for the PR. If you find it useful, I'm happy to merge it. However, it does seem a big confusing to me because you can't do an `import "mathlive";` unless you use some bundler, and also from your example it's not clear where that JavaScript file should go. Should you create a separate JavaScript file? Put it in a `<script>` section in the HTML?

hsellik:
I'm not familiar with the user base of this library, but I think the majority of the people who use npm are also using some kind of bundler. If they are not, then their project is not very serious and they are better off with the unpkg version?

Feel free to make any changes in the documentation when you see fit or start receiving issues.

PR #2239, opened 2024-01-01
Only pay attention to focusout of math-elements

dcbradley:
In commit 645fa9d66b8a63 for #1914, the math virtual keyboard was prevented from disappearing when a math-field with mathVirtualKeyboardPolicy = "manual" loses focus.  However, if subsequently something else loses focus that isn't a math-field, the virtual keyboard disappears anyway.  The attached patch avoids that problem.

Thanks for this amazing library!


arnog:
Good catch! And thanks for your kind words.

PR #2269, opened 2024-01-21
Fix: 1. For \_ in \text command to produce expected output. 2. For Generting correct when \text used with superscript/subscript 

theapu:
1. Expected output for `\text{A\_2}` is `A_2`. Current mathlive version produces output `A\_2`.
    _ is a special character in TeX and for getting character '_' one should use `\_`
2. \text{A}_2 produces incorrect mathml. current mathml output is
```
<mrow><mtext >A</mtext><msub><mn>2</mn></msub></mrow>
``` 
Expected output is
```
<msub><mtext >A</mtext><mn>2</mn></msub>
```
This pull request adds `'\\_': 0x005f` to `TEXT_SYMBOLS` in  src/latex-commands/definitions-utils.ts to fix 1.
function `scanTeX` modified in src/formats/atom-to-math-ml.ts to take care of superscript and subscripts with `\text`

### DIFF
--- a/src/formats/atom-to-math-ml.ts
+++ b/src/formats/atom-to-math-ml.ts
@@ -238,6 +238,7 @@ function parseSubsup(base: string, stream: MathMLStream, options): boolean {
   if (!superscript && !subscript) return false;
 
   let mathML = '';
+
   if (superscript && subscript)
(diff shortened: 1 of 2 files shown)


arnog:

PR #2274, opened 2024-01-25
Fix: When LaTeX array environment is used with @{} in column specifiers browser hangs.

theapu:
When array environment with `@{}` in column specifier as below the browser sometimes hangs and content always disappear or does not update in WYSIWYG editor

```
\left\{
\begin{array}{@{}l@{}}
A + B\\
C + D
\end{array}
\right\}
```  

`@{}` is commonly used in array to remove space before/after a column. If above LaTeX code is used in the mathlive demo page

Following error appears in console

```
Uncaught TypeError: Cannot read properties of undefined (reading 'shift')
    at Qp (mathlive.min.mjs:192:15305)
    at eh (mathlive.min.mjs:192:15993)
    at new U (mathlive.min.mjs:192:17147)
    at Dh (mathlive.min.mjs:2803:18674)
    at t.render (mathlive.min.mjs:2803:14292)
    at nh (mathlive.min.mjs:192:42110)
    at t.createBox (mathlive.min.mjs:192:29931)
    at t.render (mathlive.min.mjs:2803:23269)
    at nh (mathlive.min.mjs:192:42110)
    at t.createBox (mathlive.min.mjs:192:29931)
```

A check added in `getVListChildrenAndDepth` function to check if prevChild is null. If so `[null, 0]` is returned to prevent this issue.

### DIFF
--- a/src/core/v-box.ts
+++ b/src/core/v-box.ts
@@ -58,6 +58,9 @@ function getVListChildrenAndDepth(
   if ('individualShift' in params) {
     const oldChildren = params.individualShift;
     let prevChild = oldChildren[0];
+    if (prevChild == null)
+      return [null, 0];
+    
     const children: (VBoxChild | VBoxElementAndShift)[] = [prevChild];
 
     // Add in kerns to the list of params.children to get each element to be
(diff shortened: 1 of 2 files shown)


arnog:
Looks good.

Issue #2294, opened 2024-02-09
Cursor before placeholder regression


mgreminger:
## Description

This is a regression due to the most recent commit (2553750c5c9a2b471efdc7f592f37a708a5821cb). The cursor is able to be placed immediately ahead of the placeholder which allows items to be inserted before the placeholder, which can be confusing for the user. In previous versions of MathLive, the cursor could not be placed directly before the placeholder. The placeholder itself would be highlighted instead.

### Steps to Reproduce

Type the following key sequence: `/` + `RightArrow` + `a`

### Actual Behavior

The `a` will appear before the placeholder in the denominator of the fraction (see screen recording below).

https://github.com/arnog/mathlive/assets/6439649/e55ee44f-a097-4cca-9c84-5434032be249

### Expected Behavior

The `a` will replace the placeholder in the denominator of the fraction (see screen recording below of behavior before the most recent commit 2553750c5c9a2b471efdc7f592f37a708a5821cb).

https://github.com/arnog/mathlive/assets/6439649/dba2138b-80b6-4b58-8bbe-a1b2851657a0

### Environment

> Is this a regression: did it use to work in a previous version?

This is a regression, previous versions of MathLive including 0.98.6 did not have this issue.

**MathLive version**  
Most recent commit on master (2553750c5c9a2b471efdc7f592f37a708a5821cb)

**Operating System**
Windows and Linux

**Browser** 
Firefox and Chrome

Issue #2293, opened 2024-02-09
Multiplication Keyboard Shortcut Regression


mgreminger:
## Description

Thanks for the fix for #2291, it solves the issue and is much appreciated. I think this issue is related. There is a regression since 0.98.6 for inserting a multiplication symbol in the numerator of a fraction when the denominator is a placeholder. The multiplication symbol jumps to the denominator of the fraction instead of appearing in the numerator where expected.

### Steps to Reproduce

Type the following sequence into the smoke test at the most recent revision (2553750c5c9a2b471efdc7f592f37a708a5821cb):  `/` + `f` + `*` + `g`

### Actual Behavior

The multiplication symbol appears in the denominator even though the cursor was in the numerator (see screen recording below).

https://github.com/arnog/mathlive/assets/6439649/556bb76d-3b27-4c08-a333-757b402274c1

### Expected Behavior

Both the multiplication symbol and the g should appear in the numerator (see the screen recording using 0.98.6 below):

https://github.com/arnog/mathlive/assets/6439649/61a040af-cfda-4f98-8daf-0818138c3497

### Environment

> Is this a regression: did it use to work in a previous version?

This is a regression, the behavior was correct in 0.98.6

**MathLive version**  
Issue occurs with most the most recent commit on the master branch (2553750c5c9a2b471efdc7f592f37a708a5821cb)

**Operating System**
Windows and Linux

**Browser**
Firefox and Chrome

Issue #2291, opened 2024-02-08
Fraction keyboard entry regression


mgreminger:
## Description

Regression in fraction behavior since 0.98.6 (tested on 24081c25e2a0332b351bb340ecd59802df2520c9)

### Steps to Reproduce

Create a fraction using the / key and then enter the numerator and denominator. Use Tab or Arrow to get out of the fraction and type the multiplication symbol to the right of the fraction.

### Actual Behavior

The contents of the fraction will disappear and be replaced by prompts. See screen recording below:

https://github.com/arnog/mathlive/assets/6439649/b64a01f9-9d84-473e-b473-ec445757cef2

### Expected Behavior

This worked correctly in the most recent release 0.98.6. See the screen recording below:

https://github.com/arnog/mathlive/assets/6439649/852d9577-9602-4ba3-9611-fd21af9abb4f

### Environment

This is a regression, issue seen at head of master (24081c25e2a0332b351bb340ecd59802df2520c9), works correctly in 0.98.6

**Operating System** Windows and Linux

**Browser** Firefox and Chrome

arnog:
I think it's a problem with keyboard shortcuts. What keyboard keys did you type for the multiplication symbol?

mgreminger:
I used the asterisk key `*` to create the multiplication.

arnog:
OK, I've narrowed down where I think the problem might be, but I can't trigger it by pressing the `*` key. I can only trigger it by pressing `xx` and it only reverts the denominator, not the numerator. Are you trying this with the smoke test? Do you have any additional configurations to the mathfield?

PR #2295, opened 2024-02-10
tests: add regression test that covers #2291, #2293, #2294

mgreminger:
Many thanks for the fixes for #2291, #2293, and #2294. The tests for my app are now passing. I'm looking forward to rolling out the new context menu feature!

This pull request adds a Playwright regression test that covers #2291, #2293, and #2294.


arnog:
Excellent. Thanks for the PR!

PR #2299, opened 2024-02-18
fix: test for the existence of TouchEvent before using

mgreminger:
The desktop versions of Safari and Firefox do not define the `TouchEvent` constructor (see the [TouchEvent MDN page](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent#browser_compatibility) for details). Currently, two of the functions in `utils.ts` use `TouchEvent`, which leads to an unhandled `ReferenceError: Can't find variable: TouchEvent` in certain situations. In particular, I was noticing this issue when using arrows to navigate MathLive's context menu on desktop Safari leading to the arrow keys not doing anything with the context menu. This fix checks for the existence of `TouchEvent` before using it and fixes the arrow key navigation of MathLive's context menu on desktop Safari and Firefox.

### DIFF
--- a/src/ui/events/utils.ts
+++ b/src/ui/events/utils.ts
@@ -6,7 +6,7 @@ export function eventLocation(
   if (evt instanceof MouseEvent || evt instanceof PointerEvent)
     return { x: evt.clientX, y: evt.clientY };
 
-  if (evt instanceof TouchEvent) {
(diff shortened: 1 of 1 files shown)


arnog:
Ah, yes, good catch. Thanks for the PR!

PR #2326, opened 2024-03-17
fix: Safari context menu cut fix

mgreminger:
Fixes context menu Cut functionality for Safari (previously the Cut option from the context menu would do nothing on Safari). Tests have also been added for both keyboard cut/paste shortcuts and also for context menu cut/paste.

The fix checks the return value for `execCommand`, which will be `false` if it fails or is not allowed. In this case, it falls back to manually triggering the cut event.

### DIFF
--- a/src/editor-mathfield/commands.ts
+++ b/src/editor-mathfield/commands.ts
@@ -112,11 +112,12 @@ registerCommand(
       if (mathfield.model.selectionIsCollapsed) mathfield.select();
 
       if (
-        'queryCommandSupported' in document &&
-        document.queryCommandSupported('copy')
-      )
-        document.execCommand('copy');
-      else {
+        !(
+          'queryCommandSupported' in document &&
+          document.queryCommandSupported('copy') &&
+          document.execCommand('copy')
+        )
+      ) {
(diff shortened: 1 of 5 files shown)


mgreminger:
I'll need to look into why the tests are failing in CI. They pass when when I run them locally on Mac. I should have time to look into it early this week.

arnog:
Great idea. Strange that the tests are failing. Environment must be different. Let me know what you find.

mgreminger:
I was able to get the tests working. I actually needed to skip the webkit tests that used keyboard paste during CI since keyboard paste doesn't work when running headless on Linux. Strangely enough, the tests pass in GUI mode on Linux. It's not uncommon for webkit tests to fail on Linux (there are a couple of other tests that are skipped). Most importantly, the webkit tests pass on MacOS so everything is working properly on Safari.

mgreminger:
@arnog should be ready to merge now, I added the same change to the `copyToClipboard` command that I originally added to the `cutToClipboard` command since I was running into the same Safari issue in my app for copy as can be seen in the mathlive demo for cut. The mathlive context menu doesn't use `copyToClipboard`, so the Safari copy issue doesn't show up in the mathlive demo.

arnog:
Excellent. Thank you!

PR #2352, opened 2024-04-15
Add maxMatrixCols property so that matrices with more than 10 columns may be defined

mgreminger:
Currently, the matrix environment only allows 10 columns. This default makes sense since this is the default used by the amsmath library. The amsmath library provides the `MaxMatrixCols` counter to increase this limit. This pull request adds a `maxMatrixCols` property to MathLive fields to allow more than 10 columns in the matrix environment. This property is also exposed as the `max-matrix-cols` attribute.

Sorry about the large diff for `CHANGELOG.md`, I only added the new feature to the top, the linter went to town on the rest of it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,56 +1,64 @@
 ## [Unreleased]
 
+### New Features
+
+- The `maxMatrixCols` property has been added that specifies the maximum number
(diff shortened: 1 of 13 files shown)


arnog:
Looks good. Thanks for the PR, that's a great addition.

Issue #2454, opened 2024-07-16
LatexToAsciiMath incorrect interval notation


Ashp116:
The `MathLive.convertLatexToAsciiMath` doesn't properly resolve interval notations.

![image](https://github.com/user-attachments/assets/94acbe81-3993-45da-9097-c164530e7ce3)

Ashp116:
![image](https://github.com/user-attachments/assets/fdec3937-08a5-475a-b2c9-9cbb79cb75f8)


## Description

> (Optional) Provide a brief summary of the issue.

### Steps to Reproduce

> (Required) Provide steps that are specific and repeatable

1. [First I did this...]
2. [Then I did that...]
3. [And so on...]

### Actual Behavior

> (Required) What happens when you follow the steps above?

### Expected Behavior

> (Required) What did you expect to happen instead? 
> Converting from Latex to Ascii Math doesn't resolve [], (), & {}.


### Environment

```js
MathLive.convertLatexToAsciiMath('\left\lbrack\right\rbrack') // []
MathLive.convertLatexToAsciiMath('\left\lbrace\right\rbrace') // {}
MathLive.convertLatexToAsciiMath('\left(\right)') // ()
```

> Is this a regression: did it use to work in a previous version? Not sure

**Operating System** Windows

**Browser** Chrome

PR #2455, opened 2024-07-17
LatexToAsciiMath incorrect interval notation

Ashp116:
I fixed the issue #2454, where {}, [], and () wouldn't accurately convert from LaTeX to Ascii.


arnog (approved):
Thanks for the PR!
- In general avoid making unnecessary changes to the indentation or formatting of the code. Running Prettier should auto-format the code.
- There are a couple of console.log() that should be removed.
- In SPECIAL_OPERATORS, line 193 of parse-math-string.ts, the `\\{` and `\\}` entries should not be necessary, since these tokens will not appear in a ASCII Math string. However, removing lines 238-241 would take care of this case.
- Thank you for including some test cases!

PR #2496, opened 2024-08-17
Update l10n-strings.ts

physedo:
Croatian


arnog:
Manually merges to more-l10n-strings.ts

PR #2503, opened 2024-08-17
Update l10n-strings.ts

physedo:
Slovak


arnog:
Manually merged to more-l10n-strings.ts

Issue #2337, opened 2024-04-05
[Feature] Allow disabling math field focus when browser tab regains focus


pwawrzynczyk:
## Description

I'd like to be able to stop the math field from regaining focus and scrolling back to it when the browser tab is focused again. I realize the current behavior may still be desirable for some use cases, so a flag to configure it would be perfect.

### Steps to Reproduce

0. Have a page with a math field at the top and some other content below, so that if we scroll to the bottom of the page, the math field won't be visible anymore.
1. Focus the math field.
2. Scroll to the bottom.
3. Switch tab.
4. Switch back to the original tab.

### Actual Behavior

When the user switches back to the original tab, page refocuses the input and scrolls to it. This may confuse some people, who left the focus on the field, but then started looking at the rest of the page and don't want to get back to it.

### Expected Behavior

When the user switches back to the original tab, there is no scroll or focus change - the math field stays blurred.

### Environment

**MathLive version**: 0.92.1 and 0.98.6.

**Checked on**: Firefox, Chrome, Safari on desktop, Brave on mobile.

arnog:
I am not able to reproduce this. When I switch back to the original tab, the mathfield is not scrolled into view. Whether the focus is restored or not is browser dependent (Chrome does restore the focus, but Safari does not).

Do you have some event handlers perhaps that could get triggered and cause this to happen?

d7ark:
Hey @arnog. Sorry for bothering you. When could I expect this to be released?

I see this being merged 2 weeks ago, but recent release notes does not mention this. I'm not very good at figuring things like this out, so sorry if this is stupid question :D 

Have a nice day!

PR #2528, opened 2024-09-25
fix: improved refocus behavior on window exit and then return

mgreminger:
This pull request fixes two issues with focus handling when switching away from and back to a tab with a `mathfield` element to make the behavior the same as a standard HTML `input` or `textarea` element. The current behavior is correct when switching tabs (the first test in the two videos below) but doesn't correctly handle the case when switching non-overlapping windows (second test in the two videos below). This was corrected by looking for the window `blur` event rather than the `visibilitychange` event since `visibilitychange` is not triggered when switching windows that are not overlapping.

The second issue that is addressed is correctly handling the case where the user clicks off of the `mathfield` element but doesn't select another element (third test in the videos below, users commonly do this to unselect a `mathfield`). Previously, the mathfield would incorrectly regain focus when the tab is refocused even though the mathfield wasn't previously focused. This behavior is confusing to users and may cause unintentional scrolling if the user has scrolled the mathfield out of view (this may be related to the issues raised in #2337). By using the `click` event of document to cancel the refocus, this issue is corrected and now matches the normal HTML `input` element behavior.

This fix has been tested on Edge, Firefox, and Safari.

Previous focus behavior:
Previous focus behavior screen capture video

New focus behavior:
New focus behavior screen capture video

### DIFF
--- a/src/editor-mathfield/mathfield-private.ts
+++ b/src/editor-mathfield/mathfield-private.ts
@@ -1716,26 +1716,20 @@ If you are using Vue, this may be because you are using the runtime-only build o
       // don't do that for custom elements, so we do it ourselves. @futureweb
       //
 
-      // Wait for the window/document visibility to change
+      // Wait for the window focus to change
       // (the mathfield gets blurred before the window)
       const controller = new AbortController();
       const signal = controller.signal;
-      document.addEventListener(
-        'visibilitychange',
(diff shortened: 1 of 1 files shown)


arnog:
Thanks for investigating this issue and providing a PR. LGTM. I'll investigate the Playwright failures.

PR #2531, opened 2024-09-29
Add typescript definitions to ssr export

stephtr:
When importing from `'mathlive/ssr'`, typescript definitions are missing, even though there would be a fitting definition close by.


arnog:
Oops... Good catch. Thanks for the PR!

Issue #2363, opened 2024-04-30
Is it possible to "lock" the submenu of the variant?


EvanFung:
## Description
Is it possible to "lock" the submenu of the variant? After I long press the keycap, the submenu of the variant appears. When I release the left mouse button, the submenu of the variant disappears. However, is it possible to lock the submenu so that it only disappears when the user loses focus on the keyboard's submenu? Now when I first open the keyboard, long pressing shows the variant's submenu, but when I open it for a second time, I can "lock" the submenu. I don't know if locking the submenu behavior is a bug or not because when I looked at official documentation demos, I could only open submenus by long pressing on keycaps, but it will disappear when I release the mouse(I think this may be the normal behavior, but I want to lock the submenu.)

### Steps to Reproduce
1. The first time, I opened the keyboard, and pressed submenu by long press the virtual keycap, but it will disappear when I release my mouse. 
2. The second time, I opened the keyboard, and long pressed the submenu, the submenu will "lock"

### Expected Behavior
How can I lock the submenu like opening the keyboard for the second time?

### Environment
**MathLive version** 
0.68.1

**Operating System** 
MacOS

https://github.com/arnog/mathlive/assets/22959692/496cbf1c-4b41-48cc-b2b7-949cd2347f6d

**Browser**
Chrome

PR #2554, opened 2024-11-27
feat: Added option for sticky display of variants panel with a single click

### LINKED ISSUE #2549: Feature Request: Variants panel that can be persistently displayed with a single click
related #2363 

Currently, the variants panel requires a long press to display, which can be challenging for first-time users to discover. It would be helpful to have a feature that allows the panel to be persistently displayed with a single click, enabling easier selection.

I’ve implemented a poc for this feature, which you can find here:
https://github.com/arnog/mathlive/compare/master…konbu310:mathlive:poc-variants-panel-popover

https://github.com/user-attachments/assets/3169c08b-841b-466c-9359-329f0faa1a5c

Thank you for creating and maintaining such an excellent library! It has been incredibly helpful.

konbu310:
fix #2549 

### Overview

This PR adds a `stickyVariantPanel` option to the VirtualKeyboardKeycap. When this option is enabled, clicking on a key with variants configured will display the Variants Panel without a long press. Additionally, the panel will not automatically hide.

The motivation for implementing this feature is described in #2549, and a similar request can be found in #2363.

### Notes

- Tests have not been added yet.
- Added an example: sticky-variant-panel/index.html.
- I am not entirely confident about the API naming.

### DIFF
--- a/src/public/virtual-keyboard.ts
+++ b/src/public/virtual-keyboard.ts
@@ -98,6 +98,9 @@ export interface VirtualKeyboardKeycap {
 
   /** Name of the layer to shift to when the key is pressed */
   layer: string;
+
+  /** Open variants panel without long press and does not close automatically */
+  popover: boolean;
 }
 /**
  * @category Virtual Keyboard
--- a/src/public/virtual-keyboard.ts
+++ b/src/public/virtual-keyboard.ts
@@ -100,7 +100,7 @@ export interface VirtualKeyboardKeycap {
   layer: string;
 
   /** Open variants panel without long press and does not close automatically */
-  popover: boolean;
+  stickyVariantPanel: boolean;
 }
 /**
  * @category Virtual Keyboard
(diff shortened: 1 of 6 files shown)


src/public/virtual-keyboard.ts
```diff
@@ -98,6 +98,9 @@ export interface VirtualKeyboardKeycap {
 
   /** Name of the layer to shift to when the key is pressed */
   layer: string;
+
+  /** Open variants panel without long press and does not close automatically */
+  popover: boolean;
```
arnog, line 103: I would probably go with something a bit more descriptive. Maybe `stickyVariantPanel`?

konbu310: That's a good name.

src/virtual-keyboard/utils.ts
```diff
@@ -1135,19 +1135,21 @@ function handlePointerDown(ev: PointerEvent) {
   }
 
   if (keycap.variants) {
-    if (pressAndHoldTimer) clearTimeout(pressAndHoldTimer);
```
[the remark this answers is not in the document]
konbu310: It seems like I mistakenly removed it, so I’ll restore it.

arnog (approved):
Looks good. Thanks!

Issue #2550, opened 2024-11-22
Error in new Next JS 15


noams24:
I tried to migrate my app from Next js 14 to 15, and got error when compiling the app:

![image](https://github.com/user-attachments/assets/b6c85026-6b97-4d15-8c08-e9a1f9642499)

⨯ ./node_modules/mathlive/package.json
Error parsing package.json file
invalid exports field value "./dist/fonts/*" for key "./fonts/": folder exports must end with "/"

victorvianaom:
Same for me:

![image](https://github.com/user-attachments/assets/296b32dc-07bf-4be3-a4d5-c6ca559994c1)

victorvianaom:
Opened a PR to fix this #2555

PR #2555, opened 2024-11-29
Fix package.json subpath exports matching

victorvianaom:
This PR updates subpath exports in the main package.json to comply with Node.js's [subpath patterns](https://nodejs.org/api/packages.html#subpath-patterns), which utilize "direct static matching and replacement."

As a result, it resolves issue #2550, where Mathlive could not be used with Next.js when Turbopack was enabled due to its strict code validation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,8 +46,8 @@
     "./static.css": "./dist/mathlive-static.css",
     "./fonts/*": "./dist/fonts/*",
     "./sounds/*": "./dist/sounds/*",
-    "./fonts/": "./dist/fonts/*",
-    "./sounds/": "./dist/sounds/*"
+    "./fonts/": "./dist/fonts/",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -46,8 +46,8 @@
     "./static.css": "./dist/mathlive-static.css",
     "./fonts/*": "./dist/fonts/*",
     "./sounds/*": "./dist/sounds/*",
-    "./fonts/": "./dist/fonts/*",
```
victorvianaom, line 49: Or could we just remove these 2 lines?

package.json
```diff
@@ -46,8 +46,8 @@
...
-    "./sounds/": "./dist/sounds/*"
```
victorvianaom, line 50: Or could we just remove these 2 lines?

arnog:
Great! Thanks for the PR.

Issue #2530, opened 2024-09-29
`convertAsciiMathToLatex` converting fractions only if the numerator is in brackets.


stephtr:
## Description

> According to the docs, `convertAsciiMathToLatex("1/2") should return `"\\frac{1}{2}"`. Currently, it only converts it to a frac if the numerator is in brackets, eg inputting `"(1)/2"`.

### Steps to Reproduce

> (Required) Provide steps that are specific and repeatable

```JS
import { convertAsciiMathToLatex } from 'mathlive';
console.log(convertAsciiMathToLatex('1/2'));
```

### Actual Behavior

> Returns '1/2'.

### Expected Behavior

<img width="707" alt="image" src="https://github.com/user-attachments/assets/bf9f57d6-4225-4280-9c29-895ad6ae5e81">

### Environment

**MathLive version**  0.101.0

**Operating System** macOS

**Browser** Edge

PR #2556, opened 2024-12-02
FIX: Resolved the issue #2530

Ashp116:
There was an issue when you tried to convert the math ASCII string (`1/2`) to latex. You got `1/2`, which is incorrect.


arnog (commented):
I don't understand how this could fix the issue. When I try the code, it returns the same result as before (i.e `1/2`).
Did you have some other changes perhaps that are not in this PR?

PR #2560, opened 2024-12-04
Fix prompt box shifting issue in chrome browser

Seancheey:
Fixed the prompt box shifting upwards issue in chrome browser when focused


Seancheey:
Closing since it's not compatible with SSR

PR #2566, opened 2024-12-13
Only show focused prompt box when focused

Seancheey:
Fixed the issue that upon blur prompt box, there was still a surrounding outline box

Fixed result:

https://github.com/user-attachments/assets/bc148bf0-801a-4edc-9acc-66d9eba5eadf


arnog:

Issue #2254, opened 2024-01-12
VirtualKeyboard stops accepting keytouches after a few rapid keypresses


DZiems:
## Description

Typing keys in rapid succession on the virtual keyboard seems to disable/halt any further input. Pressing backspace quickly is a common use case affected by this.

### Steps to Reproduce

1. Open a mobile instance of a mathfield and focus it to bring up the virtual keyboard.
2. Type any key or keys in rapid succession.
3. After about two or three rapid keytouches, input stops being received.

### Actual Behavior

Keypresses stop inputing values.

### Expected Behavior

Keypresses should continue to input values regardless of typing speed.

### Environment

> Is this a regression: did it use to work in a previous version?

Most likely not a regression.

**MathLive version**  
0.98.5
Using the following commit from this repo (most up-to-date version as of writing).
[49f11e2] 

**Operating System** _macOS, Windows, iOS. Include the version_
iOS 17.1.2

**Browser**
Safari and Chrome have this identical behavior.

Abhishek-Nair-135:
I'm facing this issue as well. But was able to reproduce only on iOS devices.

tomas-sipko:
Having the same issue, are there any updates?

haypho:
I am also running into this issue. I pulled the repo and did some initial investigation. It seems like the the mathfield is getting blurred when typing into the virtual keyboard too quickly. I'm still doing more investigation, but thought I would post a comment here in case someone has some tribal knowledge that could assist quicker.

Here is the line that is blurring the mathfield:
https://github.com/arnog/mathlive/blob/master/src/public/mathfield-element.ts#L1931

haypho:
@arnog, any chance you can review #2567 for us?
The following commands passed locally for me:
* `npm run build`
* `npm run build production`
* `npm test`

haypho:
Hey @arnog, I'm not sure if you get the PR notifications, so I am reposting here.
I updated the PR with the recommended changes (thanks again!). I read through the documentation for contributors and I didn't see any clear next steps for myself. Are there next steps that I need to complete once #2567 is merged?

haypho:
@arnog Thanks for merging this in for us! Is this going to be patched in the current version or are we going to have to wait for the year end release? Maybe the better question is when should we expect this to be published to npm?

arnog:
This was fixed in 0.105.1

PR #2567, opened 2024-12-18
fix(virtual-keyboard): ignore blur event on touch devices

haypho:
When using the virtual keyboard on mobile, typing
quickly would cause a blur event to be dispatched.
This change ignores the blur event in the mathfield
when a touch device is recognized and the virtual
keyboard is visible.

refs: #2254

### DIFF
--- a/src/public/mathfield-element.ts
+++ b/src/public/mathfield-element.ts
@@ -1922,13 +1922,15 @@ import "https://unpkg.com/@cortex-js/compute-engine?module";
     // Ignore blur events if the scrim is open (case where the variant panel
     // is open), or if we're in an iFrame on a touch device (see #2350).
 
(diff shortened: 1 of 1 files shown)


arnog (changes requested):
I would combine this check with the following one:
```ts
if (evt.type !== 'blur') return;

const touch = isTouchCapable();

if (touch && window?.mathVirtualKeyboard?.visible) return;
if (Scrim.scrim?.state !== 'closed' || (touch && isInIframe())) return;

this._mathfield?.blur();
```

haypho:
@arnog, once this is merged, do I need to do anything on my end before a patch is published to the npm registry?

arnog:
@haypho thanks for the PR. Nothing more to do on your end. This will get incorporated with the next publish to npm.

PR #2571, opened 2024-12-23
fixHidePopoverFail

1245040330:
Fixed the issue on poor computers where the message "This might have been the result of the "beforetoggle" event handler changing the state of this popover." appears.

image
This line reports an error, causing the subsequent code to not execute

aa0c9459ac583068b9a20189a31cc4d

### DIFF
--- a/src/ui/menu/menu-list.ts
+++ b/src/ui/menu/menu-list.ts
@@ -434,7 +434,13 @@ export class _MenuListState implements MenuListState {
     // Notify our parent
     if (this.parentMenu) this.parentMenu.openSubmenu = null;
 
-    if (supportPopover() && this._element?.popover) this.element.hidePopover();
+    if (supportPopover() && this._element?.popover) {
+      try {
+        this.element.hidePopover()
+      } catch (error) {
+        
+      };
+    }
 
     // We're going to do some focus manipulation, but we don't want parents
     // to react to these events (they may think the host has lost focus and


arnog:
Do you know how to reproduce this problem?

1245040330:
https://github.com/user-attachments/assets/47575f63-d947-4d16-8032-17a54b60a1c8

electron": "^25.0.1"
Chrome/114.0.5735.289

arnog:
I don’t doubt it happens but can you share the steps to make it happen? I’d like to debug this and understand why an error is thrown. I don’t think that just catching it is enough.

1245040330:
请 clone 这个项目 "https://github.com/1245040330/electronmathlive"

1245040330:
I think the cause of this bug is a problem with the browser kernel. Currently, catching this exception can run normally under this browser kernel.

arnog:
I have cloned the project and followed the instruction. After `npm start` I get an error `sh: chcp: command not found` in the console.

Also, I don't see any use of a mathfield in the source code. Perhaps there's something missing?

1245040330:
处理 chcp报错  Add the path C:\WINDOWS\system32 under the system variable PATH; 
This project is a browser. Running mathlive in this browser will result in an error.

arnog:
OK, I've checked in a fix: 6edeaca6a6aef375b53a0c5e65886c7e5b545278

1245040330:
OK thanks

Issue #2574, opened 2025-01-01
Wrong rendering output for \coloneq operator


Flydroid:
## Description

I noticed that the `\coloneq` operator gets rendered differently than its input  `:=` . 
To showcase the issue it was necessary to show images of latex renders.

### Steps to Reproduce

Enter: `x:=5`

### Actual Behavior

Which results in latex `$$ x\coloneq5 $$` and the redering is: 
![image](https://github.com/user-attachments/assets/2c4c5b33-6d03-4a4a-b786-711f2728ef4c)

### Expected Behavior

The input results in the same latex but the rendering should be: 
![image](https://github.com/user-attachments/assets/664f3cb4-053f-4e44-b9ce-5520c1bb7857)
in accordance to the mathtool package: https://ftpmirror1.infania.net/mirror/CTAN/macros/latex/contrib/mathtools/mathtools.pdf

### Environment
Mathlive version:  0.103.0 on https://cortexjs.io/mathlive/demo/ 

### Solution

I think I can provide a pull request with a fix, but would require some guides.
The `\coloneq` operator is defined here: https://github.com/arnog/mathlive/blob/c1ea3067410212737d79f85e4aea92d65d0d9c11/src/latex-commands/definitions-utils.ts#L301
The mathtool packages now shows `\coloneqq` as a legacy.
It would propose to change the line 301 to the same as for `\coloneqq`:
`coloneq: '{\\mathop{\\char"2254}}', // ≔`      
Not sure though if more of the mathtool package code should be updated.

arnog:
Yep, you got the fix. Worth also looking at p. 31 of the mathtools package to double check other changes and make the definitions in MathLive are consistent.

PR #2575, opened 2025-01-05
Fixed \coloneq symbol in definitions-utils.ts

Flydroid:
The issues was described in https://github.com/arnog/mathlive/issues/2574

Changes:
-fixed \colonq symbol to use the character `2245` which is the correct equivalent to `:=` (as used on the legacy  \coloneqq according to the current mathtools package: https://ftpmirror1.infania.net/mirror/CTAN/macros/latex/contrib/mathtools/mathtools.pdf


arnog:
Looks good. Thanks.

PR #2591, opened 2025-02-18
[FIX]: incorrect label for psi on virtual keyboard

### LINKED ISSUE #2590: [BUG] Psi key is incorrectly labeled as zeta on the virtual keyboard
- [x] Checked for duplicate issue

**Description**

I noticed the psi symbol is labeled as zeta in the virtual keyboard on the website.

Here's a screenshot, where the psi symbol is labeled as zeta

Image

Clicking the button creates the correct `/psi` code though. Also, the capital version is correct.

I think I found the line causing it, I'll look at opening a PR.

cjgunnar:
Fixes #2590

As shown there, here's the bug:

414047585-ca7ad38c-1dd3-481d-a9d7-09ba743aad12

And here's the fixed screenshot!

image

Just one line changed 👍

### DIFF
--- a/src/virtual-keyboard/data.ts
+++ b/src/virtual-keyboard/data.ts
@@ -314,7 +314,7 @@ export const LAYOUTS: Partial<
           label: '<i>&psi;</i>',
           class: 'MLK__tex hide-shift',
           insert: '\\psi',
-          aside: 'zeta',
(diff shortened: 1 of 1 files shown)


arnog:
Excellent. Thank you!

PR #2621, opened 2025-03-10
fix: add virtual keyboard commands to the types

Mathias-S:
[`toggleVirtualKeyboard` is documented here](https://cortexjs.io/mathfield/guides/commands#user-interface), but trying to call `mf.executeCommand('toggleVirtualKeyboard')` results in type errors.

This change allows users to use `mf.executeCommand('toggleVirtualKeyboard')`, and the related `showVirtualKeyboard` and `hideVirtualKeyboard` without getting type errors.


arnog:

PR #2622, opened 2025-03-11
docs: fix dead links to the cortexjs.io website

Mathias-S:
The website was updated a few months ago, with all MathLive URLs changing:
* `https://cortexjs.io/mathlive/` → `https://cortexjs.io/mathfield/`
* `https://cortexjs.io/docs/mathlive/` → `https://cortexjs.io/mathfield/api/`

All the old links now return 404, so this PR fixes those links. Would it be possible to also create some redirects from the old URLs?

### DIFF
--- a/CHANGELOG_2017-2023.md
+++ b/CHANGELOG_2017-2023.md
@@ -259,7 +259,7 @@ mf.selection = mf.getPromptRange(id);
 
   The menu can be customized by setting the `mf.menuItems` property of the
   mathfield. The value of this property is an array of menu items. See
-  [the documentation](https://cortexjs.io/mathlive/guides/menus/) for details.
+  [the documentation](https://cortexjs.io/mathfield/guides/menu/) for details.
 
 ### Improvements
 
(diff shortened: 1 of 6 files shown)


arnog:
Hmmm... Good catch. I'll look at adding the redirects.
I'm also considering rebranding. Right now cortexjs = mathfield/mathlive + compute engine. I'm thinking of abandoning cortexjs and moving to mathlive = mathfield + compute engine (hosted on mathlive.io). Another option would be to use another name for the "umbrella" brand... Thoughts?

Mathias-S:
Yes, I noticed that you seemed to be in the process of rebranding. I tried to avoid renaming anything to keep this PR to only fixing links. Your proposal makes a lot of sense, but I imagine that renaming the npm package will be a long process if you want to get people to start using a different package.

Maybe rebranding could be discussed in an issue for more visibility? I see that you already started a discussion about it on Discord. In any case, I don't think this PR gets in the way of a possible rebranding.

arnog:
In the meantime, I’ve added the redirects

Mathias-S:
Thanks! But the README also links to https://cortexjs.io/docs/mathlive which still returns a 404.

PR #2637, opened 2025-03-21
Incorrect order of conditions for short-circuting.

winstonewert:
In my project, I had an error triggered from within MathLive on this line:

`if (this.silenceNotifications || !this.mathfield.host || !this.mathfield)`

The code is pretty clearly trying to bail out if mathfield is absent, but because the check against mathfield is last, the line errors checking mathfield.host first. I swapped the order of the checks so that mathfield is checked first.


arnog:
That's a good catch. You are quite correct. Thanks for the PR!

PR #2713, opened 2025-04-28
fix: Change the conditions for moving focus after entering a shortcut

### LINKED ISSUE #2712: Input a shortcut (Ex. cm) that is replaced by the \operatorname command, the selection becomes selected after replacement.
As the title suggests, when you enter a shortcut replaced by \operatorname such as ‘cm’, it becomes selected.

This behaviour is inconvenient because you have to press the arrow keys every time you want to continue typing.

It seems to be affected by the feature added in #2671, which moves the focus to the body after entering commands such as \sqrt.

The actual behaviour is as follows (The video was recorded on the smoke test screen using the master code cloned on 2025/04/28.)

https://github.com/user-attachments/assets/854376ee-e6f3-4afb-9f46-8466619998e5

konbu310:
fixed #2712 

### Overview

Changed to use `Atom.hasEmptyBranch(‘body’)` to determine the condition for moving focus to the body after entering a shortcut.

`\sqrt` and `\ang`, which were pointed out in #2671, now work correctly.

Other shortcuts do not seem to be affected, but please let me know if there are any problems.

https://github.com/user-attachments/assets/ff3c7124-b7be-4ede-a49f-3cbcdbfa7e23

### DIFF
--- a/src/editor-mathfield/mode-editor-math.ts
+++ b/src/editor-mathfield/mode-editor-math.ts
@@ -172,7 +172,7 @@ export class MathModeEditor extends ModeEditor {
     const data =
       typeof input === 'string'
         ? input
-        : globalThis.MathfieldElement.computeEngine?.box(input).latex ?? '';
+        : (globalThis.MathfieldElement.computeEngine?.box(input).latex ?? '');
 
     if (
       !options.silenceNotifications &&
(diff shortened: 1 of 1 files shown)


arnog:
Looks good! Thanks for the PR.

PR #2726, opened 2025-05-12
Update italian l10n-strings

GGilli:
Updated some italian localization and added localization missing keys


arnog:
I'm not sure of what exactly has changed, and why. Could you elaborate?

GGilli:
I've changed some translation issues, just an example:
'tooltip.undo': 'Disfare' -->  "tooltip.undo": "Annulla"
The problem is that "undo" in italian is translated correctly with "disfare" but this is not used in the context of a web application or program. I made a complete check of the translation of each entry matching with the context.
Also there were some entries that in the italian list were not previously evaluated like "menu.insert": "Inserisci".

Hope this helps to understand

arnog:
thanks, that helps for that. It appears that some keys were removed and some capitalization changed as well. Could you explain why that is?

Sorry for the follow up questions, but it’s really hard to tell what has changed. The diff tool shows that *everything* has changed, mostly because you changed the single quotes to double quotes.

GGilli:
Sorry about the double quotes — I probably added them during editing without noticing. What I did was take the English list and translate all the keys to make sure I didn’t miss anything.
As for capitalization, in Italian we usually only capitalize the first letter or specific proper names. In any case, it's not critical. On buttons or menu we usually find the strings both ways.

arnog:
I've merged the changes manually.

PR #2729, opened 2025-05-14
Protect resizeObserver.unobserve call

winstonewert:
In my project, I'm observing errors because mathfield.field is undefined in `mathfield.resizeObserver.unobserve(mathfield.field);` Given that the function already bails if mathfield is undefined, it seems it should do the same in mathfield.field is undefined.


winstonewert:
So I subsequently found out that this was because I was unmounting the element during the 'change' event. That was accidental, but I think mathlive probably still shouldn't throw errors

PR #2797, opened 2025-07-28
fix: create separate atoms for each column of a row

manstie:
Array.fill calls the function once for the whole array, so if the environment has multiple columns you end up with multiple cells with the same reference which creates multiple carats for input

image


ashmoreinc:
This fixes this issue: https://github.com/arnog/mathlive/issues/2800

Issue #2850, opened 2025-09-17
Unable to select superscript on multi-line math


tt-atomi:
When I create multiline math (wrapped in \displaylines) it won't let me select superscripts.
Subscripts works fine. And no issue if it's on single line with not \displaylines wrapper. 
 
<img width="286" height="210" alt="Image" src="https://github.com/user-attachments/assets/265e2486-8147-4477-908f-cacdbc480253" />

tt2166:
## Posible Cause
On multi-line. The issue was with the `nearestAtomFromPointRecursive` function in MathLive's atom selection logic. When clicking on mathematical expressions containing superscript or subscript atoms, the function would return the parent `ArrayAtom` container instead of the actual selectable atom (like the superscript or subscript).

1. `ArrayAtom` containers (which represent grouped mathematical expressions) were being returned as selectable atoms
2. The function wasn't properly searching through all children and script branches of `ArrayAtom`s
3. Script atoms (superscripts, subscripts) positioned outside the main atom bounds weren't being found

**Possible solution**
```
```javascript
// Enhanced ArrayAtom handling
if (atom instanceof ArrayAtom) {
  // Handle root ArrayAtoms (like displaylines) differently
  if (atom.isRoot) {
    for (const row of atom.rows) {
      for (const cell of row) {
        if (cell) {
          for (const atom2 of cell) {
            const r2 = nearestAtomFromPointRecursive(mathfield, cache, atom2, x, y);
            if (r2[0] <= result[0]) result = r2;
          }
        }
      }
    }
  } else {
    // Handle regular (non-root) ArrayAtoms
    for (const row of atom.rows) {
      for (const cell of row) {
        if (cell) {
          for (const atom2 of cell) {
            const r2 = nearestAtomFromPointRecursive(mathfield, cache, atom2, x, y);
            if (r2[0] <= result[0]) result = r2;
          }
        }
      }
    }
  }
  // Always search children for ArrayAtoms
  const children = atom.children;
  for (const child of children) {
    const r = nearestAtomFromPointRecursive(mathfield, cache, child, x, y);
    if (r[0] <= result[0]) result = r;
  }
}
```

PR #2861, opened 2025-09-23
Fix superscript selection in multiline math expressions (\displaylines)

tt-atomi:
Fixed issue https://github.com/arnog/mathlive/issues/2850

Problem: Clicking on superscripts in multiline math expressions wouldn't select
Solution: Removed coordinate bounds checking that prevented superscript selection in \\displaylines


arnog:
Thanks for the PR

PR #2865, opened 2025-09-30
Include all macros when creating speakable text fragments

nritschel:
Currently, if the lookup of a macro in `DEFAULT_MACROS` (for the purpose of checking if it should be expanded) fails, that macro is skipped entirely. That means that custom macros are not included in text-to-speech at all. 

I have changed the logic, so that the macro text is used as a fallback in cases where the macro cannot be looked up (e.g., a custom macro that's not in `DEFAULT_MACROS`. This provides better support for speaking custom macros until a more extensive customization feature gets added.


arnog:
Makes sense. Thanks for the PR.

PR #2866, opened 2025-10-03
fix: corrected prompt vertical spacing to account for padding

gamberoillecito:
This pull requests aims at solving the incorrect spacing in prompts, where the prompt contents are not vertically centered.
The first fix simply adjusts the prompt position with respect to the content, taking into account also the vertical padding.

**With first fix:**
image

**Without first fix:**
image

As we can see, after the first fix, the prompts are incorrectly spaced with respect to the fraction line. The second fix addresses this issue:

**With second fix:**
image

**Without second fix:**
image


gamberoillecito:
Hello, if this pull request ends up being merged, I would greatly appreciate a new release to go back using the official package instead of one I have built to address this issue.
Thank you for you time

arnog:
ok, I'll have a look at it, but it's going to take a while because I need to build test cases to verify it.

PR #2871, opened 2025-10-05
Fix: Editor remains in LaTeX mode after deleting backslash

neysanfoo:
When you type \ (backslash) to enter LaTeX command mode and then delete it with backspace, the editor remains in LaTeX mode.

I have added a check to properly return to math mode in this case.


arnog:
Thanks for the PR. A fix using a different technique has now been implemented.

PR #2946, opened 2026-02-14
Fix virtual-keyboard-toggle focus issues

mgreminger:
Fixes two issues. First, if the `math-field` is not focused, clicking `virtual-keyboard-toggle` will give it focus and it will accept virtual keyboard input. Second, if `math-field` already has focus, clicking `virtual-keyboard-toggle` will open the virtual keyboard and the `math-field` will accept input from the virtual keyboard. Previously, both cases would require re-clicking the `math-field` element for the virtual keyboard to work properly. This is what was causing two of the playwright tests to fail.

I toggle the blur/focus state on the `virtual-keyboard-toggle`  `pointerdown` handler, mirroring the focus fixes for other event types. I'm not sure if this is the best way to fix this issue or not, but it's an improvement on the current behavior where, clicking the `virtual-keyboard-toggle` button either results in no focus or zombie focus where it appears to be focused but the virtual keyboard doesn't work.

This pull request also fixes a timing issue with another one of the tests and bumps Playwright to the latest version

### DIFF
--- a/src/editor-mathfield/mathfield-private.ts
+++ b/src/editor-mathfield/mathfield-private.ts
@@ -436,11 +436,12 @@ If you are using Vue, this may be because you are using the runtime-only build o
         else {
           window.mathVirtualKeyboard.show({ animate: true });
           window.mathVirtualKeyboard.update(makeProxy(this));
+          // Scroll after VK slide-in animation completes (0.28s)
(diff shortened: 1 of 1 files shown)


arnog:
Thanks for the PR!

PR #2956, opened 2026-03-09
Simple fix of accent relation spacing before '='

repst:
- This changes AccentAtom rendering to use box type 'ord' so accented atoms (e.g. \hat{x}) use standard inter-atom spacing before relation operators (e.g. '=').
- It also adds a test in markup.test.ts to ensure spacing before '=' for \hat{x}= matches x=.


arnog:

PR #2957, opened 2026-03-10
Fix backspace trap after deleting lone latex backslash

repst:
Fixes a caret trap in LaTeX command mode after deleting a lone backslash (related to [#2871](https://github.com/arnog/mathlive/pull/2871)).

Repro: type a\, press Backspace (deletes \), then press Backspace again. Before this change, deletion/navigation could plonk because an empty latexgroup remained. This PR removes empty latexgroup containers when delete operations remove the last LaTeX atom, so the second Backspace correctly deletes a.

Compared with #2871’s targeted “last backslash” handling, this addresses the underlying structural issue in delete flow and applies to both backward/forward delete paths. Included regression test: backspace should not trap caret in empty latex group (test/playwright-tests/physical-keyboard.spec.ts).

### DIFF
--- a/src/editor-model/delete.ts
+++ b/src/editor-model/delete.ts
@@ -415,9 +415,21 @@ export function deleteBackward(model: _Model): boolean {
         return;
       }
 
+      const targetParent = target.parent!;
       model.position = model.offsetOf(target.leftSibling);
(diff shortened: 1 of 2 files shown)


arnog:

PR #2964, opened 2026-03-12
Fix duplicate atoms created by setValue() with incomplete LaTeX

Meinzzzz:
This pull requests aims at fixing a rendering/syncing issue, where  by useing setValue() calls with incomplete LaTeX (for example \sum_{ and similar constructs) could created duplicates.
The fix changes atomIsInRange() so atoms that are already offset-in-range are considered in-range even when no non-sentinel child exists.
This ensures deletion includes partially formed atoms created from incomplete LaTeX input, preventing accumulation across sync/update cycles.
bug


arnog:
Looks good. Thanks for the PR

PR #2994, opened 2026-04-10
fix: Prevent virtual keyboard recursion in sandboxed iframe mathfields

### LINKED ISSUE #2931: Infinite recursion bug when using virtual keyboard in a cross-origin iframe
When I use virtual keyboard in a cross-origin iframe I face infinite recursion:
https://github.com/user-attachments/assets/70d916e8-15b0-44c4-9905-6d286ad7991c

The bug is reproduced in this example: 
[mathlive-keyboard-bug-test.zip](https://github.com/user-attachments/files/24494595/mathlive-keyboard-bug-test.zip)

konbu310:
fix #2931

## Summary

This PR fixes a virtual keyboard message loop for sandboxed mathfields inside cross-origin iframes.

In this setup, the iframe uses a real `VirtualKeyboard` instance instead of a proxy. When a mathfield command such as `insert` is sent through a `message` event, that same `VirtualKeyboard` instance can receive the command again and re-execute it. This causes repeated input and, in some cases, a stack overflow.

The fix is to make `VirtualKeyboard` re-handle only commands that are actually targeted at the virtual keyboard itself. Commands meant for the mathfield are left to the mathfield’s own message handler, which avoids the recursion.

## Testing

Added a Playwright regression test for a sandboxed mathfield inside a cross-origin iframe.

This test fails on current `master` with repeated input and passes with this patch.

### DIFF
--- a/src/virtual-keyboard/virtual-keyboard.ts
+++ b/src/virtual-keyboard/virtual-keyboard.ts
@@ -656,8 +656,8 @@ export class VirtualKeyboard implements VirtualKeyboardInterface, EventTarget {
       // don't handle it here (the iframe's mathfield will handle it)
       if (window === window.top && source !== window) return;
 
-      // Let the local mathfield's message listener handle non-keyboard
-      // commands to avoid re-sending them to the same window.
+      // If we're receiving our own message for a mathfield command,
+      // don't re-execute it here (the local mathfield will handle it)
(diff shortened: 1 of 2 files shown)


konbu310:
I investigated and fixed this together with Codex. Please let me know if there are any issues 🙏

arnog:
LGTM. Thanks for the PR!

konbu310:
Thanks for the merge!

PR #3001, opened 2026-04-22
Add z-index to focused color in the colorpickers.

Blibbo:
Fixes stacking problem with the neighboring colors' background, which paints over the border of the active color.

Right now, it looks like this in the website:
Screenshot 2026-04-22 190232

With the fix, it becomes like this:
Screenshot 2026-04-22 204536


arnog:

PR #3002, opened 2026-04-25
Fix virtual keyboard safe area dupe

Blibbo:
The --_padding-bottom variable already comprehends env(safe-area-inset-bottom). Adding it to the --_keyboard-height variable as well caused there to be twice as much empty space below the keyboard, in those contexts where safe areas are relevant.

Here's how it looks like in a mobile browser where the safe area is zero, because it's handled by the browser itself:
photo_2026-04-25_15-57-51
and here's how it looks like on Android WebView, where safe areas are relevant:
photo_2026-04-25_15-58-32


arnog:

PR #3003, opened 2026-04-28
Fix sync between keyboard and menu themes

Blibbo:
Specifying the theme attribute on the keyboard parent overrides the system theme. Specifying it on the mathfield does too, but only if it's "dark". Dark overrides light system theme, but not the other way around. This is because of selector specificity; the minimal change I made makes explicit light mode override more specific than the system dark mode.
This is system dark mode with explicit light mode override before:
Screenshot 2026-04-28 093739
This is after:
Screenshot 2026-04-28 092451

### DIFF
--- a/src/ui/colors/colors.less
+++ b/src/ui/colors/colors.less
@@ -1,4 +1,5 @@
-:host {
+:host,
+:host([theme='light']) {
   --primary-color: #5898ff;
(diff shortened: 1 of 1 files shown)


Blibbo:
WHOOPS, didn't see there was a PR open about this, sorry

Issue #2949, opened 2026-02-24
Placeholder misalignment (Fill-in-the-Blank)


Gryphonn:
Placeholder bolder isn't properly aligned with its content.

You can reproduce it on the main page
https://mathlive.io/mathfield/guides/fill-in-the-blank/

<img width="245" height="166" alt="Image" src="https://github.com/user-attachments/assets/7b00c1c7-5d31-491e-8cbd-3552e933e558" /></br>
the cursor is too high

<img width="269" height="186" alt="Image" src="https://github.com/user-attachments/assets/4d05f607-e946-464a-b935-b8f86726eaf1" /></br>
doesn't contain the content

<img width="513" height="97" alt="Image" src="https://github.com/user-attachments/assets/9d8c1f9e-f180-441e-8d87-4d1b75d680fb" /></br>
doesn't contain the content

<img width="409" height="72" alt="Image" src="https://github.com/user-attachments/assets/3aea3ffa-dec2-4ff5-bd1e-fb4d91cb7075" /></br>
not looking good

Is there anything that can be done to improve the situation?

gamberoillecito:
I believe this is a duplicate of #2919

Issue #2515, opened 2024-09-02
Bug: Unable to fill placeholder in vector equation


fauzanalghifary:
## Description

Unable to fill placeholder in vector equation

### Steps to Reproduce

1. Visit https://cortexjs.io/mathlive/demo/ 
2. In the LaTeX input, enter `\vec{\placeholder{}}` or `\bar{\placeholder{}}` 
3. In the equation input, try to fill the placeholder.

### Actual Behavior

- The placeholder cannot be filled with a value.

### Expected Behavior

- The placeholder should be fillable.

### Environment

MathLive version current 0.101.0

Operating System iOS 17

Browser Chrome

PR #3023, opened 2026-05-19
Fix prompt size and alignment

Scienthousiaste:
Since commit 4721c291 (presumably), the placeholders' layout appear broken when we input some content, as documented in issue https://github.com/arnog/mathlive/issues/2949.

The problem is visible on the demo page of the fill-in-the-blank feature: https://mathlive.io/mathfield/guides/fill-in-the-blank/, for example adding a fraction to a prompt breaks its layout:
image

This PR mostly reverts the prompt size updating to its state pre 4721c291, with a slightly better vertical alignment (a couple pixel of "padding bottom" are added), and (maybe problematically?), a bigger default size. The prompt seems to appear correctly, even with complex/weird content:
image

The initial commit mentions solving issue #2515. However, in the current state of the master branch, the placeholders in accent commands are not fillable. The present PR doesn't solve that, and I suppose that issue #2515 should be reopened.

### DIFF
--- a/src/atoms/prompt.ts
+++ b/src/atoms/prompt.ts
@@ -83,26 +83,10 @@ export class PromptAtom extends Atom {
       : null;
     const emptyHeight = placeholderMetrics?.height ?? context.metrics.xHeight;
     const emptyDepth = placeholderMetrics?.depth ?? emptyHeight / 2;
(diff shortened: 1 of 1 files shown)


arnog:
Looks good to me. Thanks for the PR.

PR #3029, opened 2026-06-01
Upgrade Compute Engine and Playwright

nwalters512:
This PR upgrades `@cortex-js/compute-engine` to its latest version.

While trying to run tests, I ran into https://github.com/microsoft/playwright/issues/40724, and so I upgraded Playwright as well. Doing so revealed an issue in `test/playwright-tests/physical-keyboard.spec.ts`, which I addressed. After all of this, the tests passed for me locally.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,12 +9,12 @@
       "version": "0.109.2",
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


arnog:
Thanks for the PR

PR #3032, opened 2026-06-04
feat: expose macros option on validateLatex

Wigny:
Currently, the function `validateLatex` exposed in the public API accepts no options parameter, which makes it impossible to receive custom arguments and use it to validate expressions containing user-defined commands, causing false-positive `unknown-command` errors.

This PR adds an optional `macros` parameter to `validateLatex` and wires it through to the internal parser context, just like `convertLatexToMarkup` already does, allowing the user to pass custom macros when validating expressions with user-defined commands.


arnog:

PR #3060, opened 2026-08-21
Added free-text and free-math modes (multiline latex support)

Shrookin:
## Summary

Adds two experimental multiline mathfield modes:

- `free-text`: prose-first editing with inline LaTeX escapes.
- `free-math`: math-first editing with multiline rows and inline text islands.

Both modes use a multiline root represented as an array of rows, preserving the document's line structure rather than flattening it into a single mathfield row.

## `free-text`

`free-text` is intended for document-like editing while retaining MathLive's rendering and editing engine.

- Ordinary prose is rendered with MathLive's text font.
- Newlines, tabs, and empty rows remain part of the editable document.
- The free-text root follows the host direction: RTL prose and multiline rows align to the logical start, while embedded math islands remain LTR.
- Newlines, tabs, bullets, direct text styles, inline math, and plain-text/LaTeX/MathJSON output have browser regression coverage.

## Verification

- Public TypeScript declaration check passed.
- Jest: 334 tests and 252 snapshots passed.
- Focused Chromium free-mode suite: 8 tests passed.
- Focused cross-browser free-mode suite: 22 passed with 2 expected clipboard skips.
- Development build passed.

The feature is intentionally marked experimental for maintainer review.


Shrookin:
Follow-up bidi/caret fix pushed in e867c888:\n\n- free-text now resolves base direction independently per line\n- mixed Hebrew/Latin prose uses native Unicode bidi ordering\n- math and active LaTeX runs are isolated LTR, so the visual caret stays at the logical insertion point\n- added live-typing coverage for Hebrew -> English -> Hebrew and a following backslash-command line\n\nValidation: 334 unit tests passed; all 47 Chromium physical-keyboard tests passed; all 19 applicable free-text cases passed across Chromium, Firefox, and WebKit. The combined downstream runtime is published as newnotes-0.110.0-7 (474a6fcb).

PR #3069, opened 2026-09-07
fixing text mode behaviour and adding \math command

Shrookin:
## Summary
- add a `\math{}` command that parses its body as math inside prose, as the inverse of `\text{}`
- remember whether LaTeX entry began in `text` or `free-text` and restore that prose mode after bounded or unbounded completion
- allow nested LaTeX commands and spaces inside an open bounded entry, including `\math{\forall x \land \frac{1}{2}}`
- keep math placeholders in math mode until they are filled
- publish matching runtime bundles

## Dependency
This PR is stacked on #3060 because `free-text` is introduced there. Until #3060 merges, GitHub will show its prerequisite commits in this PR; after it lands, this branch can be rebased and the diff will reduce to the text-mode behavior and `\math{}` changes.

## Verification
- `npx tsc --noEmit`
- `npx jest test --runInBand` (335 passed)
- full Chromium physical-keyboard suite (48 passed)
- focused source/watch tests in Chromium, Firefox, and WebKit (9 passed)
- focused tests repeated against the checked-in `dist/mathlive.mjs` artifact in Chromium, Firefox, and WebKit (9 passed)
- targeted ESLint and Prettier checks


arnog:
Can you expand on why the `\math{}` command is necessary, since `\text{let $\forall x \in \R$}` is already supported?

Shrookin:
Of course. `\math{}` is not necessary for expressiveness; I proposed it primarily as an ergonomic counterpart to `\text{}`.
The motivation was interactive input in text-initialized fields. In free-text, neither typing nor pasting `$...$` interprets it as math. In ordinary text mode, typing it also does not create a math region, although pasting a complete balanced expression does (which might be something unintended which, if confirmed, I am more than happy to address either with or separately from this PR). `\text{let $\forall x \in \R$}` works because the complete bounded `\text{...}` command originates in math mode and is parsed as a unit, not because `$...$` acts as an interactive mode switch in a text-initialized field.
Since `\text{}` would not be the natural entry point in an already text-initialized field, I thought \math{} would provide intuitive symmetry: users familiar with either command could readily infer the other.
That said, I overlooked the existing `\ensuremath{}` command. It already covers the intended functional use case and has the advantage of being standard LaTeX. Therefore, `\math{}` would add discoverability and symmetry, but no new capability. I am happy to remove it and keep the PR focused on restoring prose mode after bounded and unbounded LaTeX input.

